PR #60, opened 2013-02-04
Addition of a max_age param to caches.

keithamoss:
Solved a problem we had where live data is updated in a file and the cache needs to expire when the file changes.

max_age also supports a timestamp or 'expire after' object, akin to refresh_before for the seeding util.


olt:
Closing this as this has merge conflicts and no passing tests.
Please do not hesitate to nudge me in the future, if you haven't received any response. GitHub messages are too easy to overlook.

PR #138, opened 2014-03-31
Cassandra cache

umeier:
The bavarian mapping agency [(Bayerische Vermessungsverwaltung)](http://vermessung.bayern.de) uses [Apache-Cassandra](http://cassandra.apache.org/) as tilestore for the tiles used in the [BayernAtlas](http://geoportal.bayern.de/bayernatlas).

We produce tiles directly into the cache, so we had to add a "read_only"-swich to the cache to avoid storing empty tiles into the cache.


umeier:
Switched from pycassa to datastax cassandra-driver. Is Python 3-ready now.

weskamm:
Is this still a thing someone is interested in? Changes are looking good on the first sight, but we need to have the conflicts resolved at least. Otherwise we may close this PR in the near future

weskamm:
closing due to inactivity

PR #158, opened 2014-09-03
Fix debian packaging

landryb:
Hi,

ive tried the make builddeb target on 1.7.1 tag to see it fail - so i've just updated the patches and the changelog - packages for me here.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,9 @@
+mapproxy (1.7.1) unstable; urgency=low
+
+  * New upstream release.
+
+ -- Landry Breuil <breuil@craig.fr>  Wed, 03 Sep 2014 13:43:41 +0200
+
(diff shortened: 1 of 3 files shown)


landryb:
For what it's worth, make builddeb target still fails the same on 1.8.0.

landryb:
cherry-picking commit 968f144a40d7abf724a86a7f1eff948ad8a62ac4 in this PR on top of 1.8.0 git tag allows make builddeb to succeed.

olt:
I've updated the font_dir patch in 1.8.x and master branch.

landryb:
Thanks ! You could have cherrypicked my commit to keep authorship.. just sayin', not a big deal :)

olt:
Oh, I did't used your patch. I've updates the patch while building new packages for the OSGeoLive DVD.
Patching patches is no fun :)

PR #190, opened 2015-08-22
Parse tilematrix as integer from kvp request

mkofahl:
WMTS REST and KVP parsers behave diffenerent when reading tilematrix, causing an TileOutOfRange exception (KVP) in some situations while REST is still working. As tilematrix is parsed as int in `WMTS100RestTileRequest`, it should be done similar in `WMTSTileRequestParams`.

### DIFF
--- a/mapproxy/request/wmts.py
+++ b/mapproxy/request/wmts.py
@@ -60,7 +60,7 @@ class WMTSTileRequestParams(RequestParams):
     def _get_coord(self):
         x = int(self['tilecol'])
         y = int(self['tilerow'])
-        z = self['tilematrix']
+        z = int(self['tilematrix'])
         return x, y, z
     def _set_coord(self, value):
         x, y, z = value


mkofahl:
According to the spec, Table 22: Parameters in GetTile operation request, tileMatrix is of type Character String. Will look for another solution.

olt:
Yes, that would be great.
There are already some parts in MapProxy that support strings as grid level names, so that you can have a tile grid with levels like 100k, 50k, 25k, instead of 0, 1, 2. But that's not completely implemented.

It would help if you provide a configuration and request URLs that fail.

mkofahl:
I suspect, any KVP URLs to the WMTS service handler with 
* numeric tilematrix and 
* tilematrix < 10

will fail. The error occurs in `grid.limit_tile`. `self.grid_sizes` is a list with keys like `00`, `01` .. `18`. However, `z` is e.g. `6` instead of `06`. Thus, the request handler will fail arround `if z not in self.grid_sizes`.

REST URLs will work, but probably with numeric tilematrix only. In contrast to the KVP parser, TileMatrix is parsed as numeric (`r'\d+'` in `URLTemplateConverter`). Then, the cache file path is built with `%02d` in `FileCache.level_location`.

As a workaround, I modified `grid.limit_tile`, but there are probably better ideas.

```diff
Index: grid.py
===================================================================
--- /tmp/grid.py 2014-01-29 11:42:52.000000000 +0100
+++ ./grid.py 2015-08-28 13:31:43.000000000 +0200
@@ -609,7 +609,9 @@
         x, y, z = tile_coord
         if isinstance(z, basestring):
             if z not in self.grid_sizes:
-                return None
+                z = "0" + z
+                if z not in self.grid_sizes:
+                    return None
         elif z < 0 or z >= self.levels:
             return None
         grid = self.grid_sizes[z]
```

olt:
Can you provide an example configuration and example requests?
I can't change anything without having a failing test case first. There are system tests that check both KVP and REST requests with `01` level and they are working...

PR #195, opened 2015-09-18
Validation

kaiCu:
Added mapproxy config validation between validation against specs and creating ProxyConfiguration objects


olt:
squashed and merged with e1e53b11918c1085a98ff629317f81684a4f9e65

PR #197, opened 2015-09-29
Image.FASTOCTREE fails on certains images

procrastinatio:
Transparent greyscale ('LA') cannot be quantized, so do nothing.


olt:
Thanks!

PR #198, opened 2015-10-20
Fix "CacheBackendError: cannot query SQLite for more than 333 tiles"

DirkThalheim:
When using MapProxy with MBTiles/SQLite based caches, large WMS requests can issue a CacheBackendError. The proposed fix will issue multiple SQLite-queries when more than 333 tiles are requested.

### DIFF
--- a/mapproxy/cache/mbtiles.py
+++ b/mapproxy/cache/mbtiles.py
@@ -198,29 +198,35 @@ class MBTilesCache(TileCacheBase):
             # all tiles loaded or coords are None
             return True
 
-        if len(coords) > 1000:
-            # SQLite is limited to 1000 args
-            raise CacheBackendError('cannot query SQLite for more than 333 tiles')
-
         if self.supports_timestamp:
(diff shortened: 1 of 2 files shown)


olt:
Thanks for fixing that limitation!
Can you extend the test by creating >2000 tiles with dummy data (see https://github.com/DirkThalheim/mapproxy/blob/master/mapproxy/test/unit/test_cache_tile.py#L175) and make sure that all tiles are loaded with your patch?

Also: The two `if len(coords) > 999 else` conditionals are not necessary, the slice operator always returns a valid list. `while len(coords) > 0:` can be shortened to `while coords:`.

DirkThalheim:
Ok, I included your code optimizations and created a test. 
Btw. I had to add a self.cache.cleanup() to the TestMBTileCache.teardown method. Otherwise my system (Windows) couldn't delete the temporary cache after running the tests.

olt:
Thanks!

PR #199, opened 2015-10-28
An slightly improvements in dimensions support in mapProxy.

mvtm-dn:
When I has deployed some WMS-based configuration I found that the mapProxy has a very limited support of dimensions. In fact dimensions supported only in WMTS service and only with preconfigured values. So I wrote small improvements which added initial support of dimensions for WMS-based services and improve WMTS dimensions support.

This pull request contains next improvements:
- added dimension support in answer on WMS GetCapabilities requests (I changed appropriate templates);
- added some properties of dimension like 'unit', 'using current value' and so on;
- dimension values now can be obtained as a result of parsing answer on GetCapability request to designated source;
- added option to define period in seconds between dimension values updates.

To minimize changes in mapProxy  I wrote lazyList decorator (util/lazylist.py) and used it to periodically data updates.

I'm beg my pardon for  bad English

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.egg-info
 *.pyc
 *.patch
+*~
 build/
 dist/
 doc/_build
--- a/mapproxy/config/loader.py
+++ b/mapproxy/config/loader.py
@@ -1410,7 +1410,7 @@ class LayerConfiguration(ConfigurationBase):
                         continue
                     conf_sources.append(self.context.sources[source_name].source())
 
-                pass
+                continue
             else:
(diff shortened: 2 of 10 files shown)


olt:
Thanks for you work.

Please read http://mapproxy.org/docs/nightly/development.html before you continue to work on this patch.

I also suggest you to split this into multiple patches:

- Extend dimension options
- Add dimensions to WMS capabilities
- Add support for updating the dimension values

For the last point: MapProxy never makes any requests for the source capabilites in normal operation (only mapproxy-util) and I'm not sure if I'd like to change this. I think it's best to discuss this in a separate issue.

PR #210, opened 2015-12-21
arcgiscache directory layout for FileCache

DirkThalheim:
This adds support for ArcGIS exploded cache directory structures in FileCache. So it's possible to use or create ArcGIS compatible caches.

Within the cache configurateion use:
  type: file
  directory_layout: arcgis

### DIFF
--- a/doc/caches.rst
+++ b/doc/caches.rst
@@ -31,10 +31,10 @@ The following backend types are available.
 This is the default cache type and it uses a single file for each tile. Available options are:
 
 ``directory_layout``:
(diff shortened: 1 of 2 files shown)


olt:
Is there a "spec" for that? Does ArcGIS always uses upper-case? The ArcGIS tile source uses lower-case https://github.com/mapproxy/mapproxy/blob/master/mapproxy/client/tile.py#L152

Can you please add this layout to https://github.com/mapproxy/mapproxy/blob/master/doc/caches.rst

Thanks.

DirkThalheim:
You're right. I looked in the wrong place and assumed it was upper case. So I checked again our ArcGIS caches and some online documentations. I couldn't find a concrete spec, but all I could found was lower case.

The layout is added to caches.rst

olt:
Thanks.

PR #212, opened 2015-12-27
Improve riak backend

ramunasd:
- support for python 3.x
- upgraded dependencies
- documentation improvements
- faster object write operations
- request timeout same as for http request

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,7 @@ cache:
 
 install:
     - "export MAPPROXY_TEST_RIAK_HTTP=http://localhost:8098"
+    - "export MAPPROXY_TEST_RIAK_PBC=pbc://localhost:8087"
     - "pip install -r requirements-tests.txt"
 
 script: nosetests mapproxy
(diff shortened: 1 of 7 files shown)


ramunasd:
Tests fails only in travis and only during pbc requests. I tried same riak and lib versions but can't reproduce this certain error.

ramunasd:
Hi @olt, is there any chances to merge this before version 1.8.2 ?

olt:
What steps need to be done to enable the Protobuf tests?

ramunasd:
Hi @olt ,
if You want enable Protobuf tests then it needs to add additional environment variable, something like this `MAPPROXY_TEST_RIAK_PBC=pbc://localhost:8087`

olt:
I meant the Travis tests. I don't have a working Riak setup, but it would be great to have it fully tested before the next release.

ramunasd:
As You see, it fails, but not because of mapproxy. Something is wrong with travis setup.

ramunasd:
It's known issue - https://github.com/basho/riak-python-client/issues/399
Next riak-python version will handle that issue.

PR #215, opened 2016-02-05
fix: InfoQuery coordinate transformation is inverted for y axis

wilsaj:
Short version:
Looks like the coordinate transformation being done for `InfoQuery.coord` has the y bounds inverted and is getting errors in the results for the Y coordinate transformation.

Longer version:
For WMS GetFeatureInfo requests, position is queried as coordinates I and J in image coordinates based on a bbox taken from the bounds of the viewing window. The bbox is in geographic coordinates, but the position is in image coordinates where the origin (0, 0) is upper left and (WIDTH, HEIGHT) are lower right. The `srs.make_lin_transf()` function used to transform between cartesian and image coordinates expects image bounds to be defined as (0, 0, WIDTH, HEIGHT) but when called for `InfoQuery.coord`, it is being defined as (0, HEIGHT, WIDTH, 0) so the Y coordinate ends up being off.

### DIFF
--- a/mapproxy/layer.py
+++ b/mapproxy/layer.py
@@ -152,7 +152,7 @@ class InfoQuery(object):
 
     @property
     def coord(self):
-        return make_lin_transf((0, self.size[1], self.size[0], 0), self.bbox)(self.pos)
+        return make_lin_transf((0, 0, self.size[0], self.size[1]), self.bbox)(self.pos)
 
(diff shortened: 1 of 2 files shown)


grandfatha:
Please consider merging this in order to extend the WMS features referenced at the bottom. We can really use this.

PR #130, opened 2014-01-06
Add support for ArcGIS data sources.

aronbierbaum:
Adds support for connecting to ArcGIS REST servers directly. There could definitely be some improvements made, but wanted to get some initial feedback and determine interest level.

Example Configuration:
```
services:
  demo: null
  wms:
    md:
      title: MapProxy WMS Proxy

layers:
  - name: world_arcgis
    title: World Imagery
    sources: [arcgis_cache]

caches:
  arcgis_cache:
    grids: [GLOBAL_MERCATOR]
    sources: [arcgissource]

sources:
  arcgissource:
    type: arcgis
    req:
      url: http://services.arcgisonline.com/ArcGIS/rest/services/World_Imagery/MapServer
      layers: show:0
      transparent: true
```

PR #221, opened 2016-04-11
Add support for ArcGIS Image/MapServer export endpoints

iandees:
Based entirely off of https://github.com/mapproxy/mapproxy/pull/130, with some extra tests and renewed interest to see it in master 😄 .

### DIFF
--- a/doc/sources.rst
+++ b/doc/sources.rst
@@ -6,6 +6,7 @@ Sources
 MapProxy supports the following sources:
 
 - :ref:`wms_label`
+- :ref:`arcgis_label`
 - :ref:`tiles_label`
 - :ref:`mapserver_label`
 - :ref:`mapnik_label`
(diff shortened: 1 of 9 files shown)


iandees:
How do I convince `mock_httpd()` to respond with a 500? I would expect this section in [`expected_req`](https://github.com/mapproxy/mapproxy/pull/221/commits/2b5c103c286763cff873b1afbe42b197faaf775a#diff-adbd209b97500625fa77e7f571e4a53cR55) to do it, but it seems to return a 200.

olt:
`status=500` should work. But you are testing for the same tile. All `SystemTest`s are stateful: If you cache a tile in test A, then you will have access to this tile in test B.

iandees:
Ah, that makes sense, thanks!

ingenieroariel:
Travis reports just one failure, in just one python version:

```======================================================================
FAIL: mapproxy.test.unit.test_utils.TestCleanupDirectory.test_remove_some
----------------------------------------------------------------------
  File "/home/travis/virtualenv/python3.3.5/lib/python3.3/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/build/mapproxy/mapproxy/mapproxy/test/unit/test_utils.py", line 382, in test_remove_some
    assert os.path.exists(filename), filename
AssertionError: /tmp/tmpr5bg1t/foo1/foo1.txt
----------------------------------------------------------------------
```

screen shot 2016-04-28 at 7 58 12 pm

iandees:
Yep, I see that. It must be a transient test failure because it's not related to anything my code is doing.

Can you try re-running the test suites to see if it still happens?

doc/sources.rst
```diff
@@ -253,6 +254,85 @@ Full example::
...
+This describes the ArcGIS source. The only required option is ``url``. You need to set ``transparent`` to ``true`` if you want to use this source as an overlay.
+::
+
+  req:
+    url: http://example.org/ArcGIS/rest/services/Imagery/MapService
+    layers: show: 0,1
+    transparent: true
+
+``coverage``
+^^^^^^^^^^^^
```
olt: Can you refer to the WMS source documentation for the redundant parts? The MapServer docs starts with "This source is based on the WMS source and most options apply to the Mapserver source too." You can list (and link) the specific options that are identical if you like (i.e coverage, supported_srs), but please avoid duplication :)

doc/sources.rst
```diff
@@ -253,6 +254,85 @@ Full example::
       transparent: true
 
 
+.. _arcgis_label:
+
+ArcGIS REST API
+"""
+
+Use the type ``arcgis`` to for ArcGIS servers.
```
olt: Can you mention that this is for the ImageServer/MapServer REST API.

mapproxy/test/system/test_arcgis.py
```diff
@@ -0,0 +1,62 @@
...
+def teardown_module():
+    module_teardown(test_config)
+
+transp = create_tmp_image((512, 512), mode='RGBA', color=(0, 0, 0, 0))
+
+class TestArcgisSource(SystemTest):
+    config = test_config
+    def setup(self):
+        SystemTest.setup(self)
+        # Doing the string joining here to handle the case of python27 and python3 rounding floats differently
```
olt: There is a bbox_aware_query_comparator option for this: https://github.com/mapproxy/mapproxy/blob/master/mapproxy/test/system/test_tms.py#L108

mapproxy/test/system/test_arcgis.py
```diff
@@ -0,0 +1,62 @@
...
+
+def setup_module():
+    module_setup(test_config, 'arcgis.yaml')
+
+def teardown_module():
+    module_teardown(test_config)
+
+transp = create_tmp_image((512, 512), mode='RGBA', color=(0, 0, 0, 0))
+
+class TestArcgisSource(SystemTest):
```
olt: Can you add a test the uses the other request parameters (layers, etc.)?

olt:
Thanks for your work. I did a review, most comments are inline.

Not sure why this test fails, but it failed for other branches as well. So, please ignore.

One general question about ImageServer/MapServer:
Is this like WCS and WMS? Do they have the same API? Is there any difference that need to be handled (and tested)?

iandees:
Thanks for your comments. I'll address those this morning.

ESRI's ImageServer/MapServer export endpoints are in general like WMS: they take a bounding geometry (usually an envelope/box) and return an image extract of the layer for that bounds in some specified format. The difference is in the name of the parameters used to specify the bounds and the output format and projection. After adding the tests you suggest above I think the existing tests cover those differences.

iandees:
@olt are there any other tests you'd like to see? The documentation changes I made are kind of hard for me to test out at the moment. Let me know if there are any other suggestions.

ingenieroariel:
Thanks @iandees I am running a downstream project with a lot of arcgis servers that should be automatically configured. I'll test this and report back.

mapproxy/client/arcgis.py
```diff
@@ -0,0 +1,35 @@
...
+        return resp
+
+    def _query_url(self, query, format):
+        req = self.request_template.copy()
+        req.params.format = format
+        req.params.bbox = query.bbox
+        req.params.size = query.size
+        req.params.bboxSR = query.srs
+        req.params.imageSR = query.srs
+        req.params.transparent = "True"
```
olt: Is this by intention? Could you either add a comment or fix this?

olt:
I think the patch is nearly ready to be merged. Just added a small comment.

mapproxy/client/arcgis.py
iandees: Not intentional. I fixed it in 81d8e65.

olt:
Thanks for your work!

PR #222, opened 2016-04-15
arcgis directory layout for mapproxy-util export

DirkThalheim:
Added support to the mapproxy-util export script for exporting file caches in an arcgis directory layout.


olt:
Can you create a new branch for this pull request and only commit the relevant change? The export documentation also needs to mention this new type.

DirkThalheim:
Created new Pull request with affected files: #224

PR #224, opened 2016-04-29
mapproxy-util export arcgis

DirkThalheim:
Ok, here comes a new Pull Request for ArcGIS cache support for mapproxy-util export which includes only the modified files. I had to struggle a bit with Git and GitHub as I'm not really into it. Hope this will do. The documentation is also extended.


olt:
Thanks for the updated patch!

PR #240, opened 2016-07-26
KeywordList for WMTS Capabilities

tjay:
Add OWS KeywordList to WMTS


olt:
Thanks for your contribution and sorry for the really long delay.

PR #246, opened 2016-08-22
Disable tests that require network via OFFLINE_TESTS env var.

sebastic:
Skipping the tests that require network connectivity is required for the Debian package builds whose build environment doesn't have network connectivity per [Debian Policy section 4.9](https://www.debian.org/doc/debian-policy/ch-source.html#s-debianrules).

The tests are only skipped when the OFFLINE_TESTS environment variable is present.

### DIFF
--- a/mapproxy/test/system/test_util_wms_capabilities.py
+++ b/mapproxy/test/system/test_util_wms_capabilities.py
@@ -17,6 +17,7 @@ from __future__ import with_statement
 import os
 
 from nose.tools import assert_raises
+from nose.plugins.skip import SkipTest
 
 from mapproxy.client.http import HTTPClient
 from mapproxy.script.wms_capabilities import wms_capabilities_command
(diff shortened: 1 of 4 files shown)


olt:
You can run all tests in an offline mode already:

    nosetests -a '!online'

Also note that all tests that use `mock_httpd`, etc. are making requests to a local test server that is automatically started by the mock function.  They do not require network connectivity.

A large part of the test coverage relies on these `mock_httpd` tests and disabling these would make the test suite a lot less usefull.

sebastic:
It doesn't look like the nosetest arguments are an option for the Debian package builds which use `python{2.7,3.5} -m nose`. Setting `NOSE_ATTR='!online'` in the environment results in several `requests missmatch` failures (which is another spelling error to fix), because the local test server is not started automatically.

olt:
I'm closing this as `-a '!online'` is already implemented and the `OFFLINE_TESTS` env would be a duplication.

PR #247, opened 2016-08-22
Disable tag_date for reproducible builds.

sebastic:
See: https://wiki.debian.org/ReproducibleBuilds/TimestampsInPythonVersionNumbers

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -6,7 +6,7 @@ with-doctest = 1
 
 [egg_info]
 #tag_build = .dev
-tag_date = true
+tag_date = false
 
 [bdist_wheel]
 universal=1


olt:
`tag_date` is on to prevent accidental *official* releases when calling `sdist`. We call `egg_info` to get untagged releases (see https://github.com/mapproxy/mapproxy/blob/master/release.py#L27).

sebastic:
The pybuild buildsystem used for the Debian package builds doesn't use those setup targets (it calls clean, config, build & install), so I guess we'll just keep carrying the patch in the Debian package if this PR is not going to be merged.

olt:
Yes, keeping the patch would be great as we like to keep our workflow. We often create intermediate releases that should be tagged with the date.

PR #249, opened 2016-08-22
Fix 'missmatch' typo, replace with 'mismatch'.

sebastic:
The typo was noticed testing the offline tests (#246).


olt:
Your change didn't catch all `RequestsMismatchError` occurrences. Fixed and merged that. 👍

PR #209, opened 2015-12-17
Cache backend for Amazon S3

procrastinatio:
This PR adds the possiblitiy to use [Amazon S3](https://aws.amazon.com/s3/) as cache backend Mapproxy.

This type of cache is currently used by the [swiss geoportal](https://map.geo.admin.ch) for reprojecting and serving geodata for 3D mode.

The 2D pregenerated [OGC WMTS tiles](http://api3.geo.admin.ch/services/sdiservices.html#wmts) using the swiss local projection (EPSG:21781) are reprojected on the fly by Mapproxy in a projection suitable for display in 3D (EPSG:4326) and stored in an Amazon S3 bucket.

Some live examples from the region of Lugano (Ticino, southern Switzerland): [Aerial images](https://s.geo.admin.ch/68d7ac5fc9), [Geological map](https://s.geo.admin.ch/68d7ac0440), [Historical maps of 1890 A.D.](https://s.geo.admin.ch/68d7ab7c2b)

Original code from https://github.com/mapproxy/mapproxy/issues/19 by @walkermatt.

Issue #19, opened 2012-08-18
Implement a S3 Cache


walkermatt:
Hi,

I've been looking at implementing an AWS S3 Cache backend based on the FileCache, I now have a working prototype which uses boto and I'd appreciate any thoughts and feedback. The code can be found here in the s3 branch: https://github.com/walkermatt/mapproxy/tree/s3.

If you'd like to set up an S3 cache then you'll need the [boto package](http://docs.pythonboto.org/) and a cache of type 's3' and a valid bucket name. For instance to configure the demo osm_cache to use an S3 cache:

```yaml
caches:
  osm_cache:
    grids: [GLOBAL_MERCATOR, global_geodetic_sqrt2]
    sources: [osm_wms]
    cache:
      type: s3
      bucket: uniquebucketid
```

The bucket ID must be unique across all of S3 as described [here](http://docs.pythonboto.org/en/latest/s3_tut.html#creating-a-bucket) so choose something unique. At present new buckets are created in the EU zone, I'll no doubt make that configurable soon.

Even though you can store any arbitrary data in S3 against a key I've chosen to effectively store files with the keys being the effective file system path. This may allow an existing cache to be uploaded to S3 using something like s3fs or even offline import then used / maintained by the S3 cache.

BTW The original motivation for looking at an S3 cache is to set up a redundant shared cache used by several MapProxy instances all running the same configuration behind a load balancer. At present I'm using file based cache stored on EBS volumes which works well but results in duplication of the cache data. I've experimented with using s3fs with a file based cache but I've had several stability issues where the s3fs mount disappeared periodically.

Cheers,

Matt.

pthorin:
Wouldn't this be better handled as a pull request? 
I would really like s3 as a cache backend as well, but I'm unsure how up to date this feature would be since it is about 3 years since this issue was created.

walkermatt:
Hi, I've not used this in production as I think the [implementation](https://github.com/walkermatt/mapproxy/blob/s3/mapproxy/cache/s3.py) could use a little work. Feel free to test / update.

olt:
I merged an updated version.

PR #252, opened 2016-09-08
New S3 cache backend

olt:
This is pull requests finishes the work from @walkermatt (#19) and @procrastinatio (#209).

- [x] Update Boto library
- [x] Add tests with Moto mock library
- [x] Make thread-safe (remove global connections and use sessions)
- [x] Do not subclass FileCache
- [x] Further testing
- [ ] Further optimizations if necessary


requirements-tests.txt
```diff
@@ -10,3 +10,19 @@ coverage==3.7
 requests==2.0.1
 six==1.4.1
 waitress==0.8.7
+boto3==1.4.0
+boto==2.42.0
```
dalbani: I think this `boto` dependency is not used anymore.

mapproxy/config/loader.py
```diff
@@ -1578,8 +1604,12 @@ def tile_layers(self, grid_name_as_path=False):
         for cache_name in sources:
             for grid, extent, cache_source in self.context.caches[cache_name].caches():
 
-                if dimensions and not isinstance(cache_source.cache, DummyCache):
+
+                from mapproxy.cache.s3 import S3Cache
+
+                if dimensions and not isinstance(cache_source.cache, DummyCache) and not isinstance(cache_source.cache, S3Cache):
```
dalbani: Could you explain why S3 caches (and dummy caches) are treated apart?
I can't really figure out why it would be so.

mapproxy/cache/file.py
```diff
@@ -42,179 +42,20 @@ def __init__(self, cache_dir, file_ext, directory_layout='tc',
         self.cache_dir = cache_dir
         self.file_ext = file_ext
         self.link_single_color_images = link_single_color_images
```
dalbani: There's one feature which would be nice to "backport" from `FileCache`: `link_single_color_images`.
AFAIK the *REST API* of S3 doesn't provide such "symbolic link" feature, compared to the common Web endpoint which supports the aptly named "Web Page Redirect": http://docs.aws.amazon.com/AmazonS3/latest/dev/how-to-page-redirect.html
But for compatibility reason I think we can reuse the same `x-amz-website-redirect-location` property / header.
The code would check if the property is set when fetching an object from S3 and retrieve the "real" object in that case.

mapproxy/config/loader.py
```diff
@@ -1578,8 +1604,12 @@ def tile_layers(self, grid_name_as_path=False):
...
                     # caching of dimension layers is not supported yet
+                    log.debug(type(cache_source.cache).__name__)
```
olt: Thanks. Missed that.

mapproxy/config/loader.py
olt: Missed these lines. They are from the (experimental) first pull request I based by changes on.

requirements-tests.txt
olt: moto (the S3 mock library) depends on boto. The whole Flask/Jinja/Werkzeug/etc. dependency chain is required by moto as well. A production setup should only require boto3 and botocore.

mapproxy/cache/s3.py
```diff
@@ -0,0 +1,168 @@
...
+
+        extra_args = {}
+        if self.file_ext in ('jpeg', 'png'):
+            extra_args['ContentType'] = 'image/' + self.file_ext
+        with tile_buffer(tile) as buf:
+            self.conn().upload_fileobj(
+                NopCloser(buf), # upload_fileobj closes buf, wrap in NopCloser
+                self.bucket_name,
+                key,
+                ExtraArgs=extra_args)
```
olt: MapProxy doesn't create buckets on its own. The user can set the default storage class when creating the bucket on the AWS dashboard, right?

requirements-tests.txt
dalbani: Given that `moto` is already defined as a dependency just below, can't `boto` be omitted?

mapproxy/cache/s3.py
dalbani: The storage class is defined *per object*, not bucket wide.

requirements-tests.txt
olt: requirements-tests.txt should contain the whole dependency tree as this is a state that is known to work. I don't want failed tests if boto or any other sub-dependency updates with incompatible changes.

dalbani: OK, I see your point.

doc/caches.rst
```diff
@@ -348,4 +367,56 @@ You can set the ``sources`` to an empty list, if you use an existing geopackage
...
+
+``profile_name``:
+  Optional profile name for `shared credentials <http://boto3.readthedocs.io/en/latest/guide/configuration.html>`_ for this cache. Alternative methods of authentification are using the  ``AWS_ACCESS_KEY_ID`` and ``AWS_SECRET_ACCESS_KEY`` environmental variables, or by using an `IAM role <http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html>`_ when using an Amazon EC2 instance.
+  You can set the default profile with ``globals.cache.s3.profile_name``.
+
+``directory``:
+  Base directory (path) where all tiles are stored.
+
+``directory_layout``:
+  Defines the directory layout for the tiles (``12/12345/67890.png``, ``L12/R00010932/C00003039.png``, etc.).  See :ref:`cache_file` for available options. Defaults to ``tms`` (e.g. `12/12345/67890.png`).
```
panchicore: +1

panchicore:
Hi @olt , I was testing that work with `djmp` and seems to work OK. I am marking @dalbani reviews I like with +1, the rest of work looks good.

My comments here: https://github.com/cga-harvard/HHypermap/issues/205#issuecomment-248772400

doc/caches.rst
olt: I've created a new issue to track this: #257

mapproxy/cache/s3.py
olt: I've created a new issue to track this: #256

olt:
Thanks for your reviews!

mapproxy/cache/file.py
dalbani: It look the liberty to create #258 for the enhancement.

olt: Yes, thanks!

PR #261, opened 2016-09-30
Update auth docs s/start_reponse/start_response/

walkermatt:
Very minor edit but I spent a few minutes wondering what was wrong when trying the examples just now. I assume `start_reponse` was just a typo.


olt:
Yes, that's a typo. Thanks!

PR #263, opened 2016-10-16
Fix link to TMS layer capabilities on demo page

dalbani:
Given a configuration like:

```
grids:
  'EPSG:3857': {base: GLOBAL_WEBMERCATOR}
services:
  demo: {}
  tms: {use_grid_names: true}
caches:
  cache1:
    cache: ...
    format: image/png
    grids: ['EPSG:3857']
```

The link on the demo page pointing to a TMS layer capabilities is broken: the `srs` query parameter has a value of `EPSG3857` whereas it should have been `EPSG:3857`.


dalbani:
As it happens, there also seems to be an issue in the OpenLayers demo configuration as well: https://github.com/mapproxy/mapproxy/blob/master/mapproxy/service/templates/demo/tms_demo.html#L35
I made it work in my case with:
```javascript
layername: '{{layer.name}}/{{layer.grid.srs.srs_code}}'
```

olt:
Thanks for your contribution and sorry for the late response. The demo for services with `use_grid_names` was improved with #326 and it seems that this addresses the same issue. So I'm closing this.

PR #264, opened 2016-10-23
Return "Last-Modified" header in homepage response based on configuration files timestamp

dalbani:
In a very specific deployment scenario of mine, I need to be able to determine the last modified date of an instance configuration.
Here's my implementation of setting such a HTTP header, should you find it worthy of being included in the upstream code base.


weskamm:
Seems like a special usecase. Will close this now, the idea / information will still be available anyways

PR #265, opened 2016-10-23
Remove unused references to "lock_timeout" in file/S3 cache implementations

dalbani:
I am probably overlooking something obvious but it seemed to me that the `lock_timeout` parameter is simply ignored in the file and S3 cache implementations.
So I propose to remove it to improve code readability.


olt:
Thanks!

PR #266, opened 2016-10-24
Fix missing bulk_meta_tiles in congfig global spec

sdhiscocks:
Fix to remove warning when using _bulk_meta_tiles_ in global cache configuration.


olt:
Thanks!

PR #267, opened 2016-10-24
Raise LockTimeout when 'lock' status from renderd

sdhiscocks:
Allows handling of LockTimeout when seeding.

Please see mapproxy/mapproxy-renderd#1


olt:
Thanks for adding this.

PR #269, opened 2016-11-07
document the bbox coordinates order

tudorbarascu:
Hi and thanks for your very nice work.

As I'm using a reverse axis CRS for my projects I expected to give the CRS bbox order in **South, East,  North, West** but I noticed that Mapproxy always expects **East, South, West, North**.

Thought of making a pull request to document this.

Thoughts?


olt:
Thanks. Made minor changes and commited it as 83c47d39c77ade516d577652a0f3b9321f1a1eec

PR #272, opened 2016-11-25
Add support for configurable layer advertisement

dalbani:
Here's a quick implementation of a new "advertised" property for layers, as you can find in GeoServer for example.
The goal is simply to hide the layer from the WMS / WMTS / TMS capabilities document.
Before spending more time on it (i.e. for unit tests & doc), I'd like to know if this feature would be deemed useful to be integrated.

### DIFF
--- a/mapproxy/config/loader.py
+++ b/mapproxy/config/loader.py
@@ -1544,7 +1544,8 @@ class WMSLayerConfiguration(ConfigurationBase):
             layer = this_layer
         else:
             layer = WMSGroupLayer(name=self.conf.get('name'), title=self.conf.get('title'),
-                                  this=this_layer, layers=layers, md=self.conf.get('md'))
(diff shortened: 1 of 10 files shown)


olt:
What is the use-case for this? It's still possible to request the layers right?
Sounds a bit like security-by-obscurity to me :-)

dalbani:
The use case is to be able to host some data in MapProxy without showing it in the capabilities.
So, yes, it's kind of "security by obscurity".
But it's a quite common ­mechanism — think of GeoServer's "advertised" property, outside the world, the "share by link" feature of Google Docs & similar.

olt:
Ok, can you please update the documentation and add some tests? (If you are still interested in this.)

weskamm:
Although this sounds like a cool feature, we will not merge this without tests and docs. We may close this PR in the near future.

weskamm:
closing due to inactivity

PR #273, opened 2016-11-26
Fix variable name typo

dalbani:


dalbani:
Speaking of the OGC pixel size, I see that this value of `0.00028` is also mentioned in https://github.com/mapproxy/mapproxy/blob/master/mapproxy/script/scales.py#L23.
Shouldn't it rather be defined at a single place?

PR #275, opened 2016-11-28
Caches mapnik map objects per process/thread

palmerj:
Sorry create this PR again from the correct branch.

I've determined that the mapproxy mapnik source class is not caching mapnik objects during the seeding process if multi-processing is used. This means mapnik map objects are unnecessarily constructed for each set of tile renders. The current code is designed to not reuse mapnik objects across proccess/thread because of unsafe sharing of database connections and filehandles.

This PR now reuses mapnik map objects when multi-processing is used by caching mapnik map objects by process/thread to ensure safe sharing of objects. I have also removed the reuse_map_objects parameter from the mapnik source class constructor as this is no longer required.

In my testing this results in up to 500% seeding performance increase when using parallel seed processes.

### DIFF
--- a/mapproxy/config/loader.py
+++ b/mapproxy/config/loader.py
@@ -893,16 +893,9 @@ class MapnikSourceConfiguration(SourceConfiguration):
         if mapnik_api is None:
             raise ConfigurationError('Could not import Mapnik, please verify it is installed!')
 
-        if self.context.renderd:
-            # only renderd guarantees that we have a single proc/thread
-            # that accesses the same mapnik map object
-            reuse_map_objects = True
-        else:
-            reuse_map_objects = False
-
         return MapnikSource(mapfile, layers=layers, image_opts=image_opts,
             coverage=coverage, res_range=res_range, lock=lock,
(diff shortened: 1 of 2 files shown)


olt:
This will need a bit more work as it will leak map objects with servers that create new threads periodically or for each request.

palmerj:
> This will need a bit more work as it will leak map objects with servers that create new threads periodically or for each request.

Ah yes silly me.

Do you have any ideas on better ways of implementing this fix? One idea that comes to mind is implementing a constructor option to only allow the functionality of this patch when called from mapproxy-seed, and leaving the current logic for server requests.

olt:
Maybe combine the old `reuse_map_objects` with the new pid safe-guard and enable that for seeding (see `self.context.seed`)?

palmerj:
Ok thanks I will investigate more.

ArneBab:
@palmerj could the approach I used in https://github.com/mapproxy/mapproxy/pull/401 be useful for you? I got to a non-leaking state, but it still needs tests that I didn’t manage to add in the time I had.

weskamm:
Closing this now due to inactivity.

PR #277, opened 2016-11-28
Fix bug with mbtiles tiles not being stored in TMS schema format.

### LINKED ISSUE #276: MBTiles cache is not storing tiles in TMS schema format
As per the MapBox MBTiles specifications (https://github.com/mapbox/mbtiles-spec/blob/master/1.2/spec.md), the values used to locate tile should be in the TMS schema format. The current code is storing tiles in the XYZ format and not flipping the y value. e.g [mapproxy/cache/mbtiles.py](https://github.com/mapproxy/mapproxy/blob/master/mapproxy/cache/mbtiles.py#L165)

palmerj:
Fixes #276

### DIFF
--- a/mapproxy/cache/mbtiles.py
+++ b/mapproxy/cache/mbtiles.py
@@ -137,7 +137,7 @@ class MBTilesCache(TileCacheBase):
             return True
 
         return self.load_tile(tile)
-
+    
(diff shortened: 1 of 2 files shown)


palmerj:
No required

PR #283, opened 2016-12-13
Add geopackage autoconfig

jsvrcek:
Since geopackage contains all of the necessary information to configure a service it would be helpful to users to have a utility to autoconfigure a service yaml file.  This is especially true when using geopackage files where only some levels are defined (i.e. 6-9), as this requires some additional configuration to display correctly in MapProxy.  

I added the configuration to the existing autoconfiguration tool as an extra option.   For testing I used a geopackage provided via http://www.geopackage.org/#ridata ( https://portal.opengeospatial.org/files/63156). 

Usage is:
```
 mapproxy-util autoconfig --geopackage=ERDC_Whitehorse_GeoPackage.gpkg
```


CharlyUnicorn:
+1 nice idea

furthermore i think that the autoconfig via mapproxy-util should get more information from the getcapabilities, e.g. contactinformations

weskamm:
Will be tackled in https://github.com/mapproxy/mapproxy/pull/779, thanks for contribution!

PR #284, opened 2016-12-22
Pass resolution range through to WMS source from ArcGIS source.

kurtraschke:
I found that the `min_res` and `max_res` parameters had no effect on an ArcGIS source, even though they are supported for WMS sources.

This PR adds code to parse those parameters for ArcGIS sources and pass the resulting `res_range` through from `ArcGISSource` to `WMSSource`.


olt:
Thanks.

PR #289, opened 2017-02-06
Add coverage based layer clipping

ramunasd:
This is very helpful when few layers must be merged and source doesn't provides transparent images.

### DIFF
--- a/doc/coverages.rst
+++ b/doc/coverages.rst
@@ -74,6 +74,14 @@ Any polygon datasource that is supported by OGR (e.g. Shapefile, GeoJSON, PostGI
   `OGR SQL support documentation <http://www.gdal.org/ogr/ogr_sql.html>`_. If this
   option is unset, the first layer from the datasource will be used.
 
+Clipping
+--------
+.. versionadded:: 1.10.0
+
+By default MapProxy tries to get and serve full source image even if coverage only touches it.
+Clipping by coverage can be enabled by setting option ``clip: true``. In such case all non covered
+areas will be served as transparent pixels.
+
 
 Expire tiles
 """"""""""""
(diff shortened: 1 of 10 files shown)


olt:
Thanks for implementing this!

I noticed, that you changed the BandMerger API to work like the LayerMerger. The LayerMerger is initialized before each merge, the BandMerger is only initialized once in the config loader (and configured with add_ops). That is the reason, that BandMerger gets the sources passed to the `merge` call, and not by calling `add`.  Now the sources of the BandMerger will grow indefinitely. 


BTW: The failed tests are related to S3 and I just pushed a change, that hopefully should fix these errors.

ramunasd:
Thanks @olt, i will restore BandMerger

olt:
Thanks for the update!
My latest changes introduced a conflict. I will fix this while merging your changes and checking that the union/diff/intersection coverages are working as well.

ramunasd:
Thanks @olt , it took a while to solve all inconsistencies in code.
I've noticed that `mapproxy-util server-develop` throws warning about unknown property `clip`. Do You have any clue why it happens? Config spec has this property.

olt:
Hmm, I'm not getting any warnings.

Can you try the mapproxy/clip_coverage branch? https://github.com/mapproxy/mapproxy/compare/clip_coverage

I added support for the new union/diff/intersection coverages and added code to handle multipolygon results from the buffer call.

You added ``clip: true`` to the seeding example in the documentation, this was by accident right?

ramunasd:
I've tested clipping and everything looks good. Except one small bug - after refactor i left wrong return type at `mapproxy/cache/tile.py:296`. Should be `return (None, None)`

olt:
The missing parentheses? That's fine. The comma defines a tuple, not the parentheses.

ramunasd:
LGTM :+1: we can merge that

PR #298, opened 2017-03-08
Allow Compact Cache v2 (ArcGIS Server >10.2)

tjay:
Read and Write tested with ArcGIS Server 10.4.1

- [ ] documentation

- [ ] tests

- [ ] merge v1 and v2

### DIFF
--- a/mapproxy/cache/compact.py
+++ b/mapproxy/cache/compact.py
@@ -384,27 +384,115 @@ class CompactCacheV2(TileCacheBase):
             return True
         return False
 
+
+class CompactCacheV2(TileCacheBase):
+
+    supports_timestamp = False
+
+    def __init__(self, cache_dir):
+        self.lock_cache_id = 'compactcache-' + \
+            hashlib.md5(cache_dir.encode('utf-8')).hexdigest()
+        self.cache_dir = cache_dir
+
+    def _get_bundle_tiles(self, tiles):
+        bundles = {}
+        for tile in tiles:
+            x, y, z = tile.coord
+            c = x // BundleDataV2.BUNDLESIZE * BundleDataV2.BUNDLESIZE
+            r = y // BundleDataV2.BUNDLESIZE * BundleDataV2.BUNDLESIZE
+            bundle_file = os.path.join(
+                self.cache_dir, 'L%02d' % z, 'R%04xC%04x' % (r, c))
+            if not bundles.get(bundle_file):
+                bundles[bundle_file] = [tile]
+            else:
+                bundles[bundle_file].append(tile)
+        return bundles
+
+    def is_cached(self, tile):
+        if tile.source or tile.coord is None:
+            return True
+        for bundle_file, bundle_tiles in iteritems(self._get_bundle_tiles([tile])):
+            with BundleDataV2(bundle_file) as bundledata:
+                for tile in bundle_tiles:
+                    return bundledata.tile_size(tile.coord) != 0
+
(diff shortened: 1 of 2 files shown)


olt:
Thanks for your pull request.

A few remarks:
- The documentation and tests need to be updated.
- The code should be formatted with PEP8[0] in mind. Especially the whitespaces are inconsistent with the rest of MapProxy. (It's `foo = bar` and `foo(bar=42)`).
- You seem to have more knowledge about the format specification (userHeaderOffset, slackSpace, etc.). Do you have a source for this?
- The v1 and v2 formats have a lot of things in common and the implementations should share code and they should do similar things the same way[1]. You should either change the v2 code to work more similar to the v1 implementation, or update the v1 implementation if you can improve it. But please provide reasonable arguments, if you change the existing implementation as it was already tested with various datasets.

[0] https://www.python.org/dev/peps/pep-0008/
[1] Things I noticed on first review: the file locking is now hidden in filehandle, and I'm unsure about the performance as it always reads/writes the complete index (~130kb)

tjay:
- ESRI released sample Code to convert tiles to V2 bundles:
https://github.com/Esri/raster-tiles-compactcache/tree/master/sample_code
The "knowledge" is based on this sample code. 

- The Implementation reads the index once per bundle/ bundle-update by using the class variable "BUNDLE_CACHE". There are a lot of possibilities to improve performance for this cache format. The focus of this implementation is fast tile-reading and fast bulk-storage. I’m not sure about the memory footprint of BUNDLE_CACHE with large scale caches. A LRU-dict could limit this.

olt:
- Sample code: Nice, didn't know that V2 is actually "documented".
- BUNDLE_CACHE: Ok, I didn't notice that. Yes, memory should be limited with an LRU cache or something similar. The implementation is tricky and I have to look at it in more detail to see if there are no race conditions, how this behaves with multiple processes, etc.
Did you make any benchmarks to compare it to a non-caching implementation (os.stat vs. one more read)?

olt:
To be clear: I'd prefer a simple implementation that is easy to reason about, without any state that is kept between multiple requests, if there is not a huge performance improvement.

tjay:
You are completely right:

- v1bundle / v2bundle : ~ 480 req/s
- v2bundle without BUNDLE_CACHE: ~ 420 req/s 
- mbtiles: ~ 425 req/s 

The tests used 45 bundles, the python interpreter fetched ~5 MB more ram because of BUNDLE_CACHE

I think, merge the v1 and v2 impletemention would be the best approach...

FilmCoder:
What is the status of this merge request?  What needs to be done to get it merged into master?  Simply olt's remarks above, which request updated tests, as well as redo the implementation of Compact Cache v2 such that it shares more code with the already existing Compact Cache v1 implementation in MapProxy?

I'm a maintainer of a fork of mapproxy: https://gitlab.com/GitLabRGI/nsg/gpep-mapproxy

tjay:
@FilmCoder The current implementation reads flawlessly bundles created from ArcGIS Server. I tested some bundles created by this implementation in AcrGIS Server 10.4 / 10.5 too
Of course the bundles must match the service cache definition (ArcGIS: tileInfo; Maproxy: grids+cache definitions).

iJoeCollins:
@tjay Ok thanks.

olt:
FYI: I'll be working on V2 support in the next weeks.

olt:
This is now superseded by #320

PR #300, opened 2017-03-24
Update __init__.py

markusuntera:
Only image_opts.format was compared (line 167). Whole image_opts should be compared, otherwise for example source jpeg couldn't be converted to jpeg with less quality.


olt:
Unfortunately, it's not that easy.
The image_opts of a source image are not always filled with all options. There are too many cases were images would have to be re-encoding when all image_opts would be checked for equality.

I think the only viable option to force re-encoding of images would be the introduction of a re_encode option.

weskamm:
Closing this due to inactivity

PR #301, opened 2017-04-04
add missing comma in response header

ramunasd:
This fixes non working client cache due to wrong http header format. We have issues with browsers and JOSM.
Directive `s-maxage` lets to store tiles in reverse proxy caches, so it might improve overall performance for people, who share same connection to internet.


olt:
Thanks! 👍

PR #309, opened 2017-05-26
Remove sqlite temp files on clean up level cache

sdhiscocks:
In some cases temporary sqlite files may remain when deleting a sqlite
cache level, in particular WAL and shared memory files when WAL is
enabled.

Failing to delete these files raises an error when trying to recreate a
level's cache.

Information on sqlite temp files: https://www.sqlite.org/tempfiles.html


olt:
Thanks!

PR #312, opened 2017-06-07
fix incorrect bbox documentation

tudorbarascu:
Hi @olt 

I think this is how it was supposed to be, right?

Thanks for your work!


olt:
Ugh, yes. Thanks!

PR #323, opened 2017-08-17
Fix all tests on Windows

olt:
Try to successfully test on Windows with Appveyor.
Most failing tests are because files are not closed before a teardown tries to remove the file, path with ``/`` are compared, tests use ``multiprocessing``.


olt:
Tests are still failing (mostly because the tests make assumptions that are not valid on Windows like files need to be closed before they can be removed, etc).
Merging this PR before master has too many changes for a successful merge.

PR #332, opened 2017-10-19
fixed wrong coordinate system definition for EPSG:3857

adegwerth:
The coordinate system definition for geopackages contains an error. this is now fixed.


olt:
Thanks!

Issue #183, opened 2015-06-30
Default WMS version


DirkThalheim:
The WMS of MapProxy assumes Version 1.1.1 as default on requests without VERSION parameter. This is due to a better supported through WMS clients. There are still 1.3.0 clients in the wild that do not use the right axis order. However this does not conform to the WMS specs, which expect as response the highest supported version. 

Currently the default version seems to be a hard coded. It would be nice if that could be configured for the WMS.

akuegeler:
We would also like to see this, because comliance to the WMS-Spec is very important to our customers.

tpohlmann:
Our customers also demand compliance to the WMS-Spec, see chapter 6.2.4 of 06-042_OpenGIS_Web_Map_Service_WMS_Implementation_Specification.pdf:
... In response to a GetCapabilities request (for which the VERSION parameter is optional)
that does not specify a version number, the server shall respond with the highest version it supports. ..

olt:
It's open source. You can modify MapProxy to do whatever _your customers demand_. If you start a discussion on how to do this, provide a patch with updated documentation and working tests, then there is a high chance that your patch will be included in MapProxy. This will solve the issue for all of you. 

If you are unable to do this on your own, then you need to pay someone to do this for you.
Or you just have to be patient and wait till someone finds the time to do this, because software does not get developed by making demands, _+1_ or _it's important_ comments.

PR #341, opened 2017-12-17
Respond with the highest version

drnextgis:
If no version number is specified in the request, the server shall respond with the highest
version it understands and label the response accordingly [OGC 01-068r3]. Related to issue https://github.com/mapproxy/mapproxy/issues/183.


olt:
Thanks for your changes!

PR #343, opened 2017-12-19
Take into account username and password

drnextgis:
This PR makes possible to specify username and password in capabilities URLs of `mapproxy-util autoconfig` command.


olt:
Thanks!

Issue #182, opened 2015-06-09
Basic auth header not added to request


smesdaghi:
adding username:password to the url does not add basic auth header to the request. The destination url is a geoserver instance which uses geonode extension for authentication. In the example below, if I leave out the `headers: Authorization ..` section, i get a LayerNotDefined exception.
```
secure_source:
    type: wms
    wms_opts:
      version: 1.3.0
    http:
      headers:
        Authorization: Basic YWRtaW46Z2Vvc2VydmVy
    req:
      url: http://admin:geoserver@192.168.99.100/geoserver/wms?
      layers: geonode:ne_50m_admin_0_countries
```
I'm using MapProxy==1.8.0. Below is the http request going out as a result of running a seed command:
```
GET /geoserver/wms?styles=&format=image%2Fpng&height=1104&bbox=-20037508.3428,9236039.00175,-9236039.00175,20037508.3428&layers=geonode%3Ane_50m_admin_0_countries&crs=EPSG%3A3857&service=WMS&request=GetMap&width=1104&version=1.3.0 HTTP/1.1
Accept-Encoding: identity
Host: 192.168.99.100
Connection: close
User-Agent: MapProxy-1.8.0
```
Same request with `curl` (with credentials) gets the image. With `wget` I had to add the `--auth-no-challenge` flag (in addition to credentials) as it was not passing the Auth header either. Our geoserver setup allows anonymous requests but uses authorization at layer level. The problem is most likely because of our setup but you've been so thorough with mapproxy that I figured you might choose to add something like a `http.force_auth` option.

olt:
Basic Authentication is handled by Python's urllib2 and it requires a 401 response before sending the Authorization header AFAIK.

> The problem is most likely because of our setup but you've been so thorough with mapproxy that I figured you might choose to add something like a http.force_auth option.

Our resources are limited, so we add features like this only if we or one of our customers needs it.

rylincoln:
Hi, I'm trying to connect to secure WMS server and get the 401 http response. I add in the username/password to the url but mapproxy says "Could not get any sources".  Could I be adding something similar to @smesdaghi to make it work?

landryb:
It seems it's possible to override the HTTPBasicAuthHandler urllib2 class to preemptively send the basic-auth header, see for example http://stackoverflow.com/questions/2407126/python-urllib2-basic-auth-problem/24048852#24048852

PR #345, opened 2018-01-18
Support for HTTP cookies

awaterme:
Hi Oliver,
I would like to propose an enhancement: Optional cookie management for HTTP sources (disabled by default).

Background: We have to integrate a proprietary WMS as source, which has difficulties if clients do not support HTTP session management through cookies. In other words: The client (mapproxy) has to accept and pass back session cookies.

I would be glad if you would agree to support such a scenario with MapProxy and accept this contribution.

So: Could you please review the changes? Important: I know almost nothing about python, so please let me know if something is not Ok. Luckily it is quite easy to make MapProxy support cookies. 
I can also offer to adjust the documentation. But before I start working on the docs, please let me know if you can image to accept this.

With best regards,
Andreas

### DIFF
--- a/mapproxy/client/http.py
+++ b/mapproxy/client/http.py
@@ -145,7 +145,7 @@ class _URLOpenerCache(object):
             handlers.append(authhandler)
             authhandler = urllib2.HTTPDigestAuthHandler(passman)
             handlers.append(authhandler)
-            if (manage_cookies):
+            if manage_cookies:
                 cj = CookieJar()
                 handlers.append(HTTPCookieProcessor(cj))
 
--- a/mapproxy/client/http.py
+++ b/mapproxy/client/http.py
@@ -29,12 +29,15 @@ from mapproxy.compat.modules import urlparse
 
 if PY2:
     import urllib2
-    from urllib2 import URLError, HTTPError
+    from urllib2 import URLError, HTTPError, HTTPCookieProcessor
     import httplib
+    from cookielib import CookieJar
 else:
     from urllib import request as urllib2
     from urllib.error import URLError, HTTPError
+    from urllib.request import HTTPCookieProcessor
     from http import client as httplib
+    from http.cookiejar import CookieJar
 
 import socket
 import ssl
@@ -130,8 +133,8 @@ class _URLOpenerCache(object):
     def __init__(self):
         self._opener = {}
 
-    def __call__(self, ssl_ca_certs, url, username, password, insecure=False):
-        cache_key = (ssl_ca_certs, insecure)
+    def __call__(self, ssl_ca_certs, url, username, password, insecure=False, manage_cookies=False):
+        cache_key = (ssl_ca_certs, insecure, manage_cookies)
         if cache_key not in self._opener:
             handlers = []
             https_handler = build_https_handler(ssl_ca_certs, insecure)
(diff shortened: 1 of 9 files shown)


mapproxy/client/http.py
```diff
@@ -142,6 +145,9 @@ def __call__(self, ssl_ca_certs, url, username, password, insecure=False):
             handlers.append(authhandler)
             authhandler = urllib2.HTTPDigestAuthHandler(passman)
             handlers.append(authhandler)
+            if (manage_cookies):
```
olt: No need for parentheses in Python.

awaterme: Do you want me to remove them?

olt:
Yes, why not. Just two questions to understand the reason behind this:

Does the server create a new session for each new request and the number of sessions is limited?
MapProxy will still start multiple sessions, depending on your deployment, but at least after restart of the server. Does that still work for you use-case?

mapproxy/client/http.py
olt: Yes please. The other lines look good!

awaterme:
Hi,
I suppose the server can handle this behaviour. It is currently used as backend for a web portal where it maintains the sessions for many users, without bigger problems for years now. So a couple of sessions for the client "MapProxy" will not cause problems. However we are not yet in production, so time will show.
I am quite happy if you consider accepting this and I am also happy with adjusting the parentheses :-)

awaterme:
Shall I try to amend the documentation, too?

awaterme:
Hi Oliver,

I have adjusted the docs now. I have also generated the docs; its seems Ok to me. Let me know otherwise.

Please check if my explanation regarding the cookie handling is correct: It is more as guess, because I don't know under which circumstances MapProxy will create new url_openers. I suppose per process...

I also added some links to #182 regarding basic authentication because I found the docs a bit misleading.

awaterme:
Hi Oliver, are you still willing to accept this pull request?
I'd be rather glad...
Please let me know if anything is missing.
Best regards, Andreas

m-kuhn:
What is the status of this? Could this be brought back into a mergeable state by rebasing on current master and fixing up the docs?

weskamm:
closing due to inactivity

awaterme:
Continued on https://github.com/mapproxy/mapproxy/pull/482

mapproxy/config_template/base_config/full_example.yaml
```diff
@@ -425,10 +425,27 @@ sources:
     req:
       # username and password are extracted from the URL and do not show
       # up in log files
+      # ATTENTION: see https://github.com/mapproxy/mapproxy/issues/182
```
weskamm, line 428: please remove this line and move the information to the docs in the appropriate section

mapproxy/config_template/base_config/full_example.yaml
```diff
@@ -425,10 +425,27 @@ sources:
...
 
+  # WMS source that requires authentication and session management
+  # through HTTP cookies
+  session_source:
+    type: wms
+    http:
+      # Accept session cookies and forward on subsequent requests
+      manage_cookies: true
+      # Use basic auth header directly
+      # see https://github.com/mapproxy/mapproxy/issues/182
```
weskamm, line 441: Same as above, please remove this line

PR #351, opened 2018-02-01
Add support for GeoTIFF (Fixes #205) WIP

### LINKED ISSUE #205: GeoTIFF output encoding
MapProxy does not include georeferencing metadata in TIFF files. 
When the output format is `image/tiff`, giving back a fully-fledged GeoTIFF would be a nice additional service.

sanfilippopablo:
This PR adds support for returning a complete GeoTIFF file with spatial metadata when the WMS request is made using format=image/geotiff.

This is a work in progress because it doesn't have tests. The thing is that in order to assert that the image was correctly georeferenced I would need to add gdal to requirements-test.txt. Is that OK?

### DIFF
--- a/mapproxy/image/georeferencing.py
+++ b/mapproxy/image/georeferencing.py
@@ -1,6 +1,7 @@
 from osgeo import gdal, osr
 from mapproxy.proj import Proj, transform
 from mapproxy.srs import get_epsg_num
+import uuid
 
 def get_geotransform(size, bbox):
     print bbox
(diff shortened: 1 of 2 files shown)


olt:
Thanks for working on this. I'm afraid that I have two objections with your solution.

- The already encoded image is decoded and encoded again. I assume that this is quite efficient, but it it still pushing a lot of bytes unnecessarily.

- It requires GDAL as a dependency, which is huge and the installation of this non-Python dependency can be a pain. I prefer a solution that uses Pillow for that as MapProxy already uses Pillow for all image related things. Storing additional (Geo)TIFF tags should be possible since Pillow 3.

sanfilippopablo:
I didn't know Pillow supported it! There's some manipulation to be done but it should work. I'll do it this weekend. 👍

sanfilippopablo:
While making it work with Pillow is feasible, is a lot of work and I currently don't have the time. Closing for now.

PR #354, opened 2018-02-13
Hide backend url in exception

fsteggink:
When one of the backend WMS servers of one of our MapProxy servers was having problems, I noticed that MapProxy was returning service exceptions with the backend URL exposed in it. This way internal host names, port numbers, layer names, and other information can be exposed to the outer world. This is fine for testing/debugging purposes, but undesirable for production servers.

Therefor I've created a patch to hide the backend URL in the OGC service exception. It is a boolean option, named hide_exception_url, in the HTTP options. By default the URL is not hidden, in order not to break current deployments. The URLs are only being hidden when the value is set to true.

I hope this patch will also be useful for others. I've also added a unit test. I'm not sure if more tests are necessary. For instance, the ssl settings in http_opts aren't always unit tested either.

### DIFF
--- a/mapproxy/client/http.py
+++ b/mapproxy/client/http.py
@@ -160,7 +160,7 @@ create_url_opener = _URLOpenerCache()
 
 class HTTPClient(object):
     def __init__(self, url=None, username=None, password=None, insecure=False,
-                 ssl_ca_certs=None, timeout=None, headers=None):
+                 ssl_ca_certs=None, timeout=None, headers=None, hide_exception_url=False):
         self._timeout = timeout
         if url and url.startswith('https'):
             if insecure:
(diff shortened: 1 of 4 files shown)


dalbani:
It seems that this patch only concerns errors coming from WMS servers.
It would be nice to cover more backend services (e.g. TMS, WMTS).
I have a practical use case when cascading a tile server where the URL shouldn't be visible to the end user:
image

petersmythe:
Thanks @fsteggink, it works nicely for us.  Please merge this PR.

I can also confirm that it works for WMTS: 
image

petersmythe:
Don't forget to add to your mapproxy.yaml:

```
globals:
  http:
    hide_exception_url: true
```

olt:
Thanks a lot and sorry for the long delay.

I renamed the option to hide_error_details and made this the new default. The reason is also hidden, which might also leak information depending on the error (and the reason is not useful for debugging without the URL anyway). The logs will still show the full error.

petersmythe:
Wonderful, thanks.  Now we can do away with that separate fork.

PR #364, opened 2018-05-05
Support for custom S3 server (eq. DigitalOcean Spaces)

mikkokut:
This pull request adds possibility to use custom S3 server for the S3 cache. It does it by adding two additional config parameters `region_name` and `endpoint_url`. Example usage with DigitalOcean Spaces added to the documentation.


mikkokut:
Closed as the #366 has also changes from this

PR #366, opened 2018-06-07
Support for custom S3 server (eq. DigitalOcean Spaces) incl. ACL support

stvno:
Building upon #364 to support ACL when uploading tiles. Digital Ocean's default policy is to set the files private and that makes it difficult to use.
This PR adds ACL support to @mikkokut work

### DIFF
--- a/doc/caches.rst
+++ b/doc/caches.rst
@@ -434,7 +434,7 @@ You can set the ``sources`` to an empty list, if you use an existing geopackage 
 
 .. versionadded:: 1.10.0
 
-Store tiles in a `Amazon Simple Storage Service (S3) <https://aws.amazon.com/s3/>`_.
+Store tiles in a `Amazon Simple Storage Service (S3) <https://aws.amazon.com/s3/>`_ or any other S3 compatible object server.
 
 
 Requirements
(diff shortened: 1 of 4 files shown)


mapproxy/config/loader.py
```diff
@@ -1121,6 +1130,9 @@ def _s3_cache(self, grid_conf, file_ext):
             directory_layout=directory_layout,
             bucket_name=bucket_name,
             profile_name=profile_name,
+            region_name=region_name,
+            endpoint_url=endpoint_url,
+	    access_control_list=access_control_list
```
olt: There is a tab character.

doc/caches.rst
```diff
@@ -434,7 +434,7 @@ You can set the ``sources`` to an empty list, if you use an existing geopackage
 
 .. versionadded:: 1.10.0
```
olt: .. versionadded:: 1.11.0
      ``region_name``, ``endpoint_url`` and ``access_control_list``

doc/caches.rst
```diff
@@ -454,6 +454,15 @@ Available options:
...
   You can set the default profile with ``globals.cache.s3.profile_name``.
 
+``region_name``:
+  Optional name of the region. You can set the default region_name with ``globals.cache.s3.region_name``
+
+``endpoint_url``:
+  Optional endpoint_url for the S3. You can set the default endpoint_url with ``globals.cache.s3.endpoint_url``.
+
+``access_control_list``:
+  Optional access control list for the S3. You can set the default access_control_list with ``globals.cache.s3.access_control_list``.
```
olt: Is this DO specific, or does this also work for S3? Please add a comment if it is DO specific.

stvno: This is a generic S3 API option, so it works on AWS, DO and any S3-like service

mapproxy/config/loader.py
stvno: grmbl stupid editors mumble mumble

stvno:
removed the tab & added the extra version info

olt:
Hmmm, tests are still failing for Python 3 because of some indention issues (https://travis-ci.org/mapproxy/mapproxy/jobs/389591246#L2922)
The Travis tests should pass, you can ignore the appveyor tests.

stvno:
I've opened the files in a proper editor and AFAIK there are no tabs left. I also updated the documentation with an example for access_control_list

mikkokut:
@stvno  I think I can close my pull request as this has also my changes? Should the description be updated to cover also the original feature?

stvno:
@mikkokut yeah I forked your code, so it is all in there, thanks for doing the hard bits :)

olt:
Merged. Thanks @stvno and @mikkokut

PR #370, opened 2018-07-19
Update wmts100capabilities.xml: ows:Keywords

tjay:
In WMTS 1.0.0 it is ows:Keywords not ows:KeywordList http://schemas.opengis.net/wmts/1.0/examples/wmtsGetCapabilities_response_RESTful.xml

What a mess.


olt:
Oh. Thanks. While you're at it, can you add dummy metadata to https://github.com/mapproxy/mapproxy/blob/master/mapproxy/test/system/fixture/wmts.yaml
The capabilities are checked against the XML schema, but they can't fail if there is no metadata.

PR #373, opened 2018-07-20
Fix spelling errors.

sebastic:
The lintian QA tool reported a spelling error for the latest Debian package build:

 * everytime -> every time


olt:
Thanks.

PR #374, opened 2018-07-22
Handle x robots tag from MapProxy - WIP - Discussion

eldk:
Hello,

I have commited this to  Handle X-Robots-tag header from MapProxy and avoid bots to cache pictures.

This is tested with Apache2 and mod_wsgi.

Some docs on X-Robots-tag usage from Google : https://developers.google.com/search/reference/robots_meta_tag?hl=fr

Thanks,

Eric

### DIFF
--- a/mapproxy/config/defaults.py
+++ b/mapproxy/config/defaults.py
@@ -93,4 +93,5 @@ http = dict(
     concurrent_requests = 0,
     method = 'AUTO',
     access_control_allow_origin = '*',
+    x_robots_tag = 'all',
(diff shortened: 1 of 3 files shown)


olt:
Thanks for your contribution, but it's too specific for inclusion in MapProxy. 

Your Apache should deliver a robots.txt. This is supported by all well behaving crawlers.
If you do need to support X-Rotbots, then you can write a custom middleware that add this header (https://stackoverflow.com/a/3859300).

PR #377, opened 2018-08-17
Support caching of dimension layers with WMS and WMTS

trondmm:
This patch adds caching of layers with dimensions for FileCache caches. Tested with WMS and WMTS (restful) clients. Example config using wms service with temporal layers is attached

It also adds support for dimensions for the WMS and WMTS demo clients

[mapproxy_dimension_demo.yaml.txt](https://github.com/mapproxy/mapproxy/files/2297491/mapproxy_dimension_demo.yaml.txt)


jingsam:
This feature is just what I want so far. I have a mutli-dimensional time WMTS source, and I want proxy it to support multi-projection WMS service. @olt Would you consider this PR?

olt:
Just had a glimpse on the patch:
- I would make cache.dimensions not a global option.
- How does it handle configurations with caches which have different dimensions (or no dimensions at all).
- This needs extensive tests and documentation.

weskamm:
Closing in favor of https://github.com/mapproxy/mapproxy/pull/449

PR #386, opened 2018-11-07
Bug Fix Add same logic as other tile sources without timestamps

jsvrcek:
This fixes a bug with continuing a seed using geopackage.  Currently trying to continue seeding a geopackage will result in a TypeError if using a progress file.  This changes the logic to match  other tile sources without timestamps (i.e. MBTiles). 

To test create a seed job using a geopackage using a progress file.  Then try to continue that seed job.


olt:
Thanks!

PR #393, opened 2019-01-19
HTTP error handler for WMS and ArcGis sources

ramunasd:
Same error handler as it is in tiled source.


olt:
Thanks!

ramunasd:
Youre welcome @olt . Any thoughts when a new version could be released? More than year passed since latest 1.11.0...

PR #401, opened 2019-03-26
Concurrent instances for mapnik

ArneBab:
This pull-request recovers caching of the mapnik object by segmenting the cache by process_id and thread_id, both for seeding and for serving. To avoid delays in serving after threads are reaped, mapnik.Map(…) objects are pregenerated for the least recently used mapfile.

The segmented cache speeds up the seeding process in our setup by at least factor 4 and pregeneration of the maps reduces the delay for serving not-yet-cached map tiles is reduced from ~5 seconds to ~1 second.

Licensing under Apache license is OK.

I’d be glad if you could merge these changes and hope they prove useful!


ArneBab:
The failing tests are due to existing commits on master which also fail the tests.

olt:
A patch to support caching of map objects is welcomed and tying the map object to the PID might be a working solution.

However, the current patch is too complex. Can you make it work without the use of globals, without background threads and preloading. Can you also extend the tests? Also make sure that the travis-ci tests pass (The mapnik tests are skipped as mapnik is not installed on travis. Bonus points if you can add mapnik to travis.yaml).

ArneBab:
Thank you for your answer!

I can easily get rid of the background threads and preloading (they are already separate from the caching change and added in the second commit). Getting rid of globals is harder, because they are how I made the cache survive being used in multiprocessing. Would the change with a cache as global variable but without the preloading (and background thread) be OK?

I saw the failing tests, but cannot tell whether they are due to my changes, because they already failed before my changes. The last commit with succeeding tests is from February 2018. https://github.com/mapproxy/mapproxy/commit/668b6202d89ac8315c79b2606a458ee3a77c35c2

Can you create a branch from this commit? Then I could rebase the pull-request against that branch to get tests run by travis.

olt:
I don't understand the multiprocessing part. The cache should not be available in multiple processes (as the Map object is only valid in the process it was instantiated).

Tests for master are working: https://travis-ci.org/mapproxy/mapproxy/builds

ArneBab:
I see — I did not inspect the test result info from github closely enough. And I see the issue — will take care of Python 2.7 compatibility.

ArneBab:
The global variable for multiprocessing does not guarantee the same cache to every process, it just ensures that new processes see the variable. Using the PID in the key ensures that two different processes never get the same cached values.

Without the preloading of mapnik instances the tests work. The queue class I used isn’t available in 2.7.

Do you prefer a reverted preloading commit or should I force-push the `concurrent-instances-for-mapnik` branch to remove the preloading change from history?

Thank you for your review!

ArneBab:
The tests pass now.

mapproxy/source/mapnik.py
```diff
@@ -93,24 +96,42 @@ def render(self, query):
...
-            with self._map_objs_lock:
-                if mapfile not in self._map_objs:
-                    m = mapnik.Map(0, 0)
-                    mapnik.load_map(m, str(mapfile))
-                    self._map_objs[mapfile] = m
-
-        return self._map_objs[mapfile]
+        # segment the cache by process and thread to avoid interference
+        if self._cache_map_obj:
+            cachekey = None # renderd guarantees that there are no concurrency issues
```
olt: All MapnikSources will re-use the same Mapnik object (with the first loaded mapfile).

mapproxy/source/mapnik.py
```diff
@@ -93,24 +96,42 @@ def render(self, query):
...
+            thread_id = threading.current_thread().ident
+            process_id = multiprocessing.current_process()._identity
+            cachekey = (process_id, thread_id, mapfile)
+        if cachekey not in _map_objs:
+            _map_objs[cachekey] = self._create_map_obj(mapfile)
+
+        # clean up no longer used cached maps
+        process_cache_keys = [k for k in _map_objs.keys()
+                              if k[0] == process_id]
+        if len(process_cache_keys) > (5 + threading.active_count()):
```
olt: Why 5? Can you add a comment to describe your intentions.

mapproxy/source/mapnik.py
```diff
@@ -93,24 +96,42 @@ def render(self, query):
...
-
-        return self._map_objs[mapfile]
+        # segment the cache by process and thread to avoid interference
+        if self._cache_map_obj:
+            cachekey = None # renderd guarantees that there are no concurrency issues
+        else:
+            thread_id = threading.current_thread().ident
+            process_id = multiprocessing.current_process()._identity
+            cachekey = (process_id, thread_id, mapfile)
+        if cachekey not in _map_objs:
```
olt: You can reuse map objects across multiple threads. Just take them out of the _map_obj during use and push them back afterwards. Make sure to use locking (threading.Lock) when accessing _map_obj. And also store the PID with the map object and verify that the PID matches the current PID.

mapproxy/source/mapnik.py
```diff
@@ -49,16 +50,18 @@
...
         MapLayer.__init__(self, image_opts=image_opts)
         self.mapfile = mapfile
         self.coverage = coverage
         self.res_range = res_range
         self.layers = set(layers) if layers else None
         self.scale_factor = scale_factor
         self.lock = lock
-        self._map_objs = {}
-        self._map_objs_lock = threading.Lock()
+        # global objects to support multiprocessing
```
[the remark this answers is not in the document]
ArneBab: The reason for the global object is that this cache persists within the same process in the process-pool. Therefore consecutive MapnikSource instances on the same worker process can share the map object.

mapproxy/source/mapnik.py
ArneBab: That would have been a bug — thank you for catching it! The mapfile should still be part of the key.

mapproxy/source/mapnik.py
ArneBab: This sounds like the best tool to use here would be a dict of queues identified by the PID, with the cleanup pushing map objects back into the queue. Does that sound about right?

mapproxy/config/loader.py
```diff
@@ -918,9 +918,13 @@ def source(self, params=None):
         else:
             reuse_map_objects = False
 
+        concurrent_tile_creators = self.context.globals.get_value('concurrent_tile_creators', self.conf,
```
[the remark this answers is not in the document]
ArneBab: that’s a leftover from the pregeneration of mapnik maps — I removed it. Thanks!

ArneBab:
> Threading and multiprocessing is hard. Please extend the tests (like accessing a MapnikSource in parallel from multiple threads) to make sure the basics are working (now and in the future).

A test fitting that could be to request map_obj() from a single instance in multiple threads and (while keeping the threads running) ensure that when the thread is different, the Map objects are never the same (`not map1 is map2`). Does that sound sufficient to you?

Could you give me a hint where to best locate these tests?

Adding mapnik for the travis-ci is likely outside scope for me (and outside budget, since I don’t have the experience with configuring CI for github, so I would have too much of experimentation to do to get finished on time).

mapproxy/source/mapnik.py
ArneBab: I now implemented queues for unused mapnik objects. I’ll still need help in adding a clean test for this.

I found test/unit/test_async.py, but I am not sure how to use it correctly to test the specific mapnik instance with threading.

mapproxy/source/mapnik.py
ArneBab: The mapfile is now part of the key.

mapproxy/source/mapnik.py
ArneBab: I added a comment which explains that 5 is purely empirical.

ArneBab:
The pushed changes adjusts putting back the mapnik objects based on the mapfile and makes the tests in the CI-system succeed.

mapproxy/source/mapnik.py
ArneBab: I now added an additional check to ensure that the PID always matches the PID of the process when retrieving a mapfile.

I see test failures for that commit, but the same test failures [currently occur on master](https://travis-ci.org/mapproxy/mapproxy/builds/559887389):
```
mapproxy/test/unit/test_cache_riak.py::TestRiakCacheHTTP::test_double_remove FAILED [ 48%]
```

ArneBab:
> Threading and multiprocessing is hard. Please extend the tests (like accessing a MapnikSource in parallel from multiple threads) to make sure the basics are working (now and in the future).

Can you (or someone from the project) support me in extending the tests? I tried adding a test, but got lost in the testing structure, so I did not find how best to start. Is there a chat you could recommend for getting help on that?

ArneBab:
A version that’s been rebased on master where all tests pass is in https://github.com/mapproxy/mapproxy/pull/447

weskamm:
ok so this can be closed than

PR #402, opened 2019-03-28
Feature Recommendation: Cache Prefetchers

Aords:
This pull request is a proof of concept looking for an opinion on whether MapProxy should provide prefetchers as a feature. If by the end of this pull request the feature seems like a good fit, I’ll gladly edit, document, and test the code before any merge into MapProxy.

What is a prefetcher: A prefetcher sits between a request and a cache in MapProxy. It modifies incoming requests to request additional tiles. A very simple prefetcher, like expander in this pull request, ensures that all tiles surrounding a requested tile are cached. More advanced prefetchers like Previous K Movements (PKM), and Retrospective Adaptive Prefetch (RAP) use heuristics to cache predicted future tiles based on previous tile requests. Many additional algorithms also exist with varying degrees of success.

Why use a prefetcher: There are several common tile prefetching algorithms that all attempt to minimize tile cache misses. Depending on workflow and source latency, prefetchers can provide the requesting client with a huge increase in responsiveness. Research shows that in common map tasks, such as navigation and web map interaction, requests are quite predictable. I simulated a prefetcher known as Global User-Driven Prefetching and Caching (GUCD) and found for a multiuser tile-cache with MapProxy I experienced a 15% speedup (I’m happy to share the data/code if you are interested).

Why should MapProxy support prefetchers: Integrating prefetchers into MapProxy would give users an easy way to improve performance under many common use cases. In addition, with the way prefetchers are currently setup in the pull request, existing configurations don’t change if a user doesn’t want any prefetchers.

Pull Request Summary:
At this stage there is one implemented prefetcher, expander. Expander’s expansion_amount determines the radius of tiles from the requested tile that should be cached when a tile request is made. A prefetcher can be configured to effect multiple caches and any cache can have multiple prefetchers. If this seems like something good for MapProxy just let me know and I’ll start writing tests, documentation, and additional prefetcher options.

Working Sample Configuration: [pastebin](https://pastebin.com/1Zimvr7F)

### DIFF
--- a/mapproxy/cache/file.py
+++ b/mapproxy/cache/file.py
@@ -103,6 +103,9 @@ class FileCache(TileCacheBase):
         Fills the `Tile.source` of the `tile` if it is cached.
         If it is not cached or if the ``.coord`` is ``None``, nothing happens.
         """
+
+        print("Requested Tile: ", tile)
+
         if not tile.is_missing():
             return True
 
--- a/mapproxy/cache/file.py
+++ b/mapproxy/cache/file.py
@@ -103,7 +103,6 @@ class FileCache(TileCacheBase):
         Fills the `Tile.source` of the `tile` if it is cached.
         If it is not cached or if the ``.coord`` is ``None``, nothing happens.
         """
-
         if not tile.is_missing():
             return True
 
--- a/mapproxy/cache/file.py
+++ b/mapproxy/cache/file.py
@@ -104,8 +104,6 @@ class FileCache(TileCacheBase):
         If it is not cached or if the ``.coord`` is ``None``, nothing happens.
         """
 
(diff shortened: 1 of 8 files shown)


olt:
Thanks for your work. However, I'm not convinced that this feature is worth the added complexity. 

The "expander" will always load 9 tiles instead of 1 tile for WMTS requests, right? The client needs to wait if one of the tiles is missing, even if it is not the requested tile.
For missing tiles, MapProxy already does some kind of "expanding" with the meta tile feature. 

Also, more advanced prefetchers are hard to implement because MapProxy uses multiple processes when run with mod_wsgi, gunicorn, etc. and these processes can't directly share state.

weskamm:

Issue #408, opened 2019-04-29
S3: Profiles not working as expected


dstozek:
## Context

Connecting to an S3 cache using a named profile does not seem to work. 

## Expected Behavior

An S3 cache configuration should support connecting with any credentials profile defined for the local user.

## Actual Behavior

Immediately after launching the dev server using `mapproxy-util`, server crashes with the following exception:
```Traceback (most recent call last):
  File "/home/stozek/mapproxy/mapproxy/cache/s3.py", line 63, in __init__
    self.bucket = self.conn().head_bucket(Bucket=bucket_name)
  File "/home/stozek/mapproxy/venv/lib/python3.5/site-packages/botocore/client.py", line 314, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/home/stozek/mapproxy/venv/lib/python3.5/site-packages/botocore/client.py", line 612, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.exceptions.ClientError: An error occurred (403) when calling the HeadBucket operation: Forbidden

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/stozek/mapproxy/venv/bin/mapproxy-util", line 11, in <module>
    load_entry_point('MapProxy', 'console_scripts', 'mapproxy-util')()
  File "/home/stozek/mapproxy/mapproxy/script/util.py", line 387, in main
    commands[command]['func'](args)
  File "/home/stozek/mapproxy/mapproxy/script/util.py", line 85, in serve_develop_command
    app = make_wsgi_app(mapproxy_conf, debug=options.debug)
  File "/home/stozek/mapproxy/mapproxy/wsgiapp.py", line 110, in make_wsgi_app
    services = conf.configured_services()
  File "/home/stozek/mapproxy/mapproxy/config/loader.py", line 211, in configured_services
    return self.services.services()
  File "/home/stozek/mapproxy/mapproxy/config/loader.py", line 1810, in services
    new_services = creator(service_conf or {})
  File "/home/stozek/mapproxy/mapproxy/config/loader.py", line 1858, in tms_service
    layers = self.tile_layers(conf, use_grid_names=use_grid_names)
  File "/home/stozek/mapproxy/mapproxy/config/loader.py", line 1829, in tile_layers
    for tile_layer in layer_conf.tile_layers(grid_name_as_path=use_grid_names):
  File "/home/stozek/mapproxy/mapproxy/util/py.py", line 78, in wrapper
[...]

PR #409, opened 2019-04-29
s3: fix profile_name not being passed properly to s3_session() (#408)

dstozek:
Fixes issues described in #408.

A new test in  would be nice in ` mapproxy/test/unit/test_cache_s3.py` but `moto` does not play nice with profiles, at least not in the version that's used by mapproxy.


olt:
Thanks!

Issue #410, opened 2019-05-17
Add alternative behaviour to file cache 'link_single_color_images': use hardlinks


tacatac:
Feature request (not a bug)

## Context
I'm running into a rather pathological issue with a file cache consuming most of the inodes available on its filesystem and the single colour image symlinks actually taking up more space on disk than image data.

Granted, this is mostly due to the fact that the filesystem has a rather large inode size set (512 bytes) and that single colour tiles outnumber regular tiles 20 to 1 (I should use coverages).

However I think using hardlinks is still a valid way to improve the useful-data-to-inode ratio compared to symlinks. Given the potential number of tiles in a cache, it's something to consider.

## Drawbacks

This would of course restrict file caches to single filesystems (if they even support hardlinks). And they may be slightly less practical to move across filesystems than relative symlinks.

As far as OS support is concerned it seems both `os.link` and `os.symlink` are available for Unixes and some Windows since Python 3.2:

https://docs.python.org/3/library/os.html#os.link
https://docs.python.org/3/library/os.html#os.symlink

I'm not sure if there is any other functionality that is used by MapProxy that would depend on symlinks rather than hardlinks.

## Implementation

Just switching to hard links is a matter of not setting a relative path for `real_tile_loc` https://github.com/mapproxy/mapproxy/blob/2317a159495dbb62c41d3c44eb1561a54c79bd83/mapproxy/cache/file.py#L167 and replacing `os.symlink` with `os.link` https://github.com/mapproxy/mapproxy/blob/2317a159495dbb62c41d3c44eb1561a54c79bd83/mapproxy/cache/file.py#L170

I'm not sure what the best way is to provide both symlink and hardlink alternatives.

Regards,

PR #411, opened 2019-05-18
Hardlink option for single color images in file caches

tacatac:
This is a possible implementation to allow hardlinks to be used, rather than symlinks, with the `link_single_color_images` option for file caches, as mentionned in #410, to reduce inode usage.

Current behaviour is preserved, i.e. setting `link_single_color_images` to `true` will result in symbolic links being used.

Setting the value to 'hardlink', however, will use hardlinks. The 'symlink' keyword is also added.

Mixing types works here by virtue of non-empty strings testing True.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -415,6 +415,12 @@ with no data (e.g. water areas, areas with no roads, etc.).
 
 If set to ``hardlink``, MapProxy will store the duplicate tiles as hard links.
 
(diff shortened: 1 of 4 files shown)


olt:
Can you extend the docs with pro and cons for this? Especially the timestamp issue?

tacatac:
I've added some information, I hope it's clearer and sorry for the delay!

As mentioned in the comment, the note on Windows compatibility might be overly pessimistic but unfortunately I don't currently have a way of testing that hypothesis.

weskamm:
Finally its time to merge :-) Thanks for the contribution!

PR #412, opened 2019-06-24
wms: add georeference to TIFF images and support tiff_compression

olt:


tacatac:
The new imports in `mapproxy/compat/image.py` require Pillow >=3.1 to work I believe.

#421 updates the dependency information in the install documentation (not the configuration documentation reference).

Regards,

Issue #415, opened 2019-08-06
Enable Riak tests on updated Travis-CI images.


olt:
Travis-CI updated all tests VMs to Xenial and these images do not contain riak services.

Manual installation/configuration of riak would allow us to re-enable the tests.

/cc @ramunasd Can you help with that?

ramunasd:
sure, np, already did this before

PR #416, opened 2019-08-06
Enable riak cache tests on travis-ci

ramunasd:
Fix for #415

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,9 @@ services:
 
 addons:
   apt:
+    sources:
+      - sourceline: 'deb https://packagecloud.io/basho/riak/ubuntu/ xenial main'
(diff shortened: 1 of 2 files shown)


olt:
Thanks. Not sure why the CI runs are not showing up here, but they did run https://travis-ci.org/mapproxy/mapproxy/pull_requests and somehow fail.

(You can ignore the Python 2.7/USE_LATEST_PILLOW=1 errors).

ramunasd:
Looks like travis-ci does not like when git push is forced and previous build is not finished.
I will trigger build later, seems everything will be ok.

ramunasd:
Ready.
Only 2.7 with latest pillow fails.

olt:
Thanks for the quick fix!

PR #420, opened 2019-09-11
srs: Make SupportedSRS iterable

sdhiscocks:
This resolves an issue with the demo service, which expects an iterable list of SRSs: https://github.com/mapproxy/mapproxy/blob/bc9e7dbcf3900e29066c927c04273da14e584fcc/mapproxy/service/demo.py#L142-L145

from WMS Source:
https://github.com/mapproxy/mapproxy/blob/bc9e7dbcf3900e29066c927c04273da14e584fcc/mapproxy/config/loader.py#L574-L580


olt:
Thanks for the fix. Closes #430

PR #421, opened 2019-09-17
doc: update Pillow dependency information

tacatac:
MapProxy commit 2608c510356ce3484b9723d34029fc92af2494ee for PR#412
changes `mapproxy/compat/image.py` to import ImageFileDirectory_v2 from
PIL.TiffImagePlugin.

This class only appeared in Pillow release 3.0.0.

TiffTags is also imported from PIL.TiffImagePlugin which itself only
imported it from release 3.1.0.


olt:
Thanks!

PR #425, opened 2019-10-08
Add no-cache http headers to error responses.

walkermatt:
Avoids caching in client and also allows upstream proxies etc. to better
handle WMS `application/vnd.ogc.se_inimage` and
`application/vnd.ogc.se_blank` error responses.

### DIFF
--- a/mapproxy/exception.py
+++ b/mapproxy/exception.py
@@ -44,11 +44,13 @@ class RequestError(Exception):
         """
         if self.request is not None:
             handler = self.request.exception_handler
-            return handler.render(self)
+            resp = handler.render(self)
         elif self.status is not None:
-            return Response(self.msg, status=self.status)
+            resp = Response(self.msg, status=self.status)
         else:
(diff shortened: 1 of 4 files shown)


walkermatt:
We've been looking at using MapProxy behind [AWS CloudFront](https://aws.amazon.com/cloudfront/). Due to the fact that WMS uses HTTP 200 even for exceptions adding cache control headers is a reliable way to ensure responses are not cached by the proxy (CloudFront) or the user agent.

We provide services via http and https which presents another challenge when it comes to capabilities  responses (WMS, WMTS and TMS). CloudFront sees http and https responses as equivalent and hence caches the first response which includes the corresponding protocol. Subsequent requests using a different protocol receive the cached response with the wrong protocol declared in the capabilities document. Would you be open to capabilities responses also having the same headers as I'm proposing for exceptions?

olt:
Thanks for the patch and updated test and sorry for the long delay.

PR #433, opened 2019-11-19
Change deprecated imports for Python 3.8 compat

lordi:
This changes a number of imports so that mapproxy can be used with Python 3.8.

Amongst others, this handles the removal of deprecated `cgi` methods: https://bugs.python.org/issue33843


olt:
Thanks!

PR #447, opened 2020-02-28
Concurrent instances for mapnik rebased

ArneBab:
This pull-request recovers caching of the mapnik object by segmenting the cache by process_id and thread_id, both for seeding and for serving. To avoid delays in serving after threads are reaped, mapnik.Map(…) objects are pregenerated for the least recently used mapfile.

The segmented cache speeds up the seeding process in our setup by at least factor 4 and pregeneration of the maps reduces the delay for serving not-yet-cached map tiles is reduced from ~5 seconds to ~1 second.

Licensing under Apache license is OK.

I’d be glad if you could merge these changes and hope they prove useful!

rebase of https://github.com/mapproxy/mapproxy/pull/401 on top of current master.

### DIFF
--- a/doc/sources.rst
+++ b/doc/sources.rst
@@ -602,6 +602,8 @@ Other options
 
 The Mapnik source also supports the ``min_res``/``max_res``/``min_scale``/``max_scale``, ``concurrent_requests``, ``seed_only`` and ``coverage`` options. See :ref:`wms_label`.
 
+Mapnik can be used in multithreading and multiprocessing operation by setting ``multithreaded: True``. This is only tested and safe *for seeding*.
+
 
 Example configuration
 ^^^^^^^^^^^^^^^^^^^^^
--- a/mapproxy/config/loader.py
+++ b/mapproxy/config/loader.py
@@ -952,9 +952,13 @@ class MapnikSourceConfiguration(SourceConfiguration):
         else:
             reuse_map_objects = False
 
+        concurrent_tile_creators = self.context.globals.get_value('concurrent_tile_creators', self.conf,
+            global_key='cache.concurrent_tile_creators')
+
         return MapnikSource(mapfile, layers=layers, image_opts=image_opts,
-            coverage=coverage, res_range=res_range, lock=lock,
-            reuse_map_objects=reuse_map_objects, scale_factor=scale_factor)
+                            coverage=coverage, res_range=res_range, lock=lock,
+                            reuse_map_objects=reuse_map_objects, scale_factor=scale_factor,
+                            concurrent_tile_creators=concurrent_tile_creators)
 
 class TileSourceConfiguration(SourceConfiguration):
     supports_meta_tiles = False
--- a/mapproxy/config/loader.py
+++ b/mapproxy/config/loader.py
@@ -952,13 +952,9 @@ class MapnikSourceConfiguration(SourceConfiguration):
         else:
             reuse_map_objects = False
 
-        concurrent_tile_creators = self.context.globals.get_value('concurrent_tile_creators', self.conf,
-            global_key='cache.concurrent_tile_creators')
-
         return MapnikSource(mapfile, layers=layers, image_opts=image_opts,
                             coverage=coverage, res_range=res_range, lock=lock,
-                            reuse_map_objects=reuse_map_objects, scale_factor=scale_factor,
-                            concurrent_tile_creators=concurrent_tile_creators)
+                            reuse_map_objects=reuse_map_objects, scale_factor=scale_factor)
 
 class TileSourceConfiguration(SourceConfiguration):
     supports_meta_tiles = False
--- a/mapproxy/config/loader.py
+++ b/mapproxy/config/loader.py
@@ -930,6 +930,7 @@ class MapnikSourceConfiguration(SourceConfiguration):
         res_range = resolution_range(self.conf)
 
         scale_factor = self.conf.get('scale_factor', None)
+        multithreaded = self.conf.get('multithreaded', False)
 
         layers = self.conf.get('layers', None)
         if isinstance(layers, string_type):
@@ -954,7 +955,8 @@ class MapnikSourceConfiguration(SourceConfiguration):
 
         return MapnikSource(mapfile, layers=layers, image_opts=image_opts,
                             coverage=coverage, res_range=res_range, lock=lock,
-                            reuse_map_objects=reuse_map_objects, scale_factor=scale_factor)
+                            reuse_map_objects=reuse_map_objects, scale_factor=scale_factor,
+                            multithreaded=multithreaded)
 
 class TileSourceConfiguration(SourceConfiguration):
     supports_meta_tiles = False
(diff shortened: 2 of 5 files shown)


ArneBab:
In https://github.com/mapproxy/mapproxy/pull/401#issuecomment-482585407 I was asked to add a test for this, but I need some support:

> A test fitting that could be to request map_obj() from a single instance in multiple threads and (while keeping the threads running) ensure that when the thread is different, the Map objects are never the same (`not map1 is map2`). Does that sound sufficient to you?
> 
> Could you give me a hint where to best locate these tests so that they fit your test structure?
> 
> Adding mapnik for the travis-ci is likely outside scope for me (and outside budget, since I don’t have the experience with configuring CI for github, so I would have too much of experimentation to do to get finished on time).

ArneBab:
This pull-request speeds up mapnik-operations by at least factor 4, and I would love to see it merged in mapproxy so others can benefit from the improvement, too.

ArneBab:
It’s been another 6 months now, so I’d like to ask for support again: Could you give me a hint where to best locate these tests so that they fit your test structure?

Adding mapnik for the travis-ci is likely outside scope for me since I don’t have the experience with configuring CI for github, so I would have too much of experimentation to do to get finished. Is there already a test that uses mapnik?

justb4:
In a recent project starting to use MP with Mapnik and ran into performance issues when serving tiles. Direct Mapnik `render_tile.py` without MP btw was also fast, multiple tiles, keeping the Mapnik object.   

So tried the `mapnik.py` from this PR by directly Docker-mounting, yes dirty, it into the MP install. A significant performance improvement! Though in my setup saw some GeoPackage locking errors. My setup:

* Docker Desktop for Mac (later Ubuntu) 4.3.1. Engine 20.10.11
* MapProxy 1.13.2 debian:buster-slim via [Docker Image](https://github.com/justb4/docker-mapproxy)
* `uwsgi 2.0.20 (64bit): MapProxy with 4 processes and 4 threads, proto=http-socket 
compiled with version: 8.3.0 on 12 January 2022 14:34:38 os: Linux-5.10.76-linuxkit #1 SMP Mon Nov 8 10:21:19 UTC 2021`
* Source: Mapnik with OpenTopoMap style, PostGIS, GeoTIFF Dem Hillshade, cache: GeoPackage
* Cache: GeoPackage

Sometimes when flooding tile requests via `uwsgi` saw Geopackage Locking errors like:
`FileNotFoundError: [Errno 2] No such file or directory: '/var/mapproxy/cache/tile_locks/gpkg53ff33c59ea47f3e3cc272035b380fee-67404-87968-17.lck'
Traceback (most recent call last):`

But never saw these errors when seeding (concurrency 8), which seemed much faster than serving via `uwsgi`. Running `seed_only` with `upscale_tiles` would be preferred. 

So I guess the errors seen could come from a complex multiprocessing/threading interplay of `uwsgi`, `Mapnik` source and GeoPackage (SQLite) caching locking.  Or maybe even Docker file-handling...Will need more time and tweaking with e.g. `uwsgi` processes and threads counts. But eventually will be running `seed_only` to max scale (18) with `upscale_tiles` as the area is just NL.  
 
Anyway I hope we can move this PR forward somehow, as its gains are significant. Maybe as a config option (`multithreaded: true|false`) , such that  the existing `mapnik.py` is not affected and we  can start with the MT version as "experimental". There are so many configs and setups that this is very hard to test.

weskamm:
Thanks @ArneBab for your contribution so far and sorry for the silence.
Do you think you could add a config option like @justb4 has mentioned above (e.g. similar to that in https://github.com/mapproxy/mapproxy/pull/401) ? 
That way one needs to explicitly activate the new behaviour and we would keep the old behaviour as default.

Regarding the tests, you can find an existing mapnik test over here (maybe it helps):
https://github.com/mapproxy/mapproxy/blob/master/mapproxy/test/system/test_mapnik.py

justb4:
Yes, I see, been using the `mapnik.py` version from this PR for months now. The locking problems I found seemed had to do with the complexity (many layers) of the Mapnik XML, complex SQL queries, Postgres optimization etc.. This has been cleared, and found no problems with the PR-version here.

Even more optimization: when developing a Mapnik style it is beneficial to immediately see results, as there are no direct tools like Tilemill here and we're not yet using CartoCSS. For this I integrated [Python Watchdog](https://pypi.org/project/watchdog/) in this PR's `mapnik.py` to listen to changes in the directory tree of Mapnik files. On change events, the Mapnik objects are reloaded. Together with special MapProxy Layers without caching (`disable_storage: true`) developing becomes quite convenient. Once this PR is done I can try to integrate my work in another PR, though I think 'auto-reloading' should be a config option and was bit tricky in an Multi-* context (like Thread-safety etc).

justb4:
@weskamm I have rights to approve the workflow and merge, but you have a better understanding of MapProxy internals. Making this feature optional and rebasing is a big step forward. Once merged, I can start PR-ing my "Hot Mapnik conf reloading" (see above and [here](https://github.com/justb4/docker-mapproxy-mapserver-mapnik/tree/main/patches) in a similar, i.e. making it optional, fashion. And @ArneBab thanks for patience!

ArneBab:
I created a third branch that’s rebased where all the workflows pass. 
https://github.com/DisyInformationssysteme/mapproxy/pull/3

Would you like me to file a PR from this here, too? Or would you prefer that I force-push into the current branch so this PR gets updated?

ArneBab:
I now force-pushed the rebased changes into  concurrent-instances-for-mapnik-rebased.

ArneBab:
@justb4 @weskamm please mention me here if you see more required changes to get this merged.

weskamm:
Ok im getting confused. Could you please point me to the current PR and close the others? Thanks!

weskamm:
so tests still missing?

ArneBab:
I did not create additional tests, because I did not get the existing mapnik tests to work on master without these changes (I spent far more time than expected trying to get a mapnik+seeding docker-setup working for testing all tests in test/system/test_mapnik.py) and they are skipped in the workflow, so I would have had to write these blind. Others here have tested the output of seeding. Over the weekend people will be running a large seeding to detect potential problems.

ArneBab:
If a test is a strict requirement (with concurrency, because that’s the only place where this can go wrong), I’ll see what I can do the next weeks, but this is a very slow process, because I have to ping-pong with those who have a full setup.

@justb4 can you check whether the speed improvements persist with the refactoring I did?

weskamm:
Test should usually be included if possible, but i understand the troubles you have here, so it should be ok. So maybe lets wait for final feedback from @justb4 and then we can merge this in.

weskamm:
I am merging this now without further feedback. As its configurable and defaults to the old behaviour, we should not expect any issues for default installations. Thanks for contribution!

ArneBab:
Thank you!

justb4:
Very sorry, missed this completely! I can only say that I am revisiting `mapnik.py` (and `geopackage.py`) due to recent upgrades and [patches to the Docker Image](https://github.com/justb4/docker-mapproxy-mapserver-mapnik/tree/main/patches). I will let know via Discussions/issues/PRs here.

ArneBab:
No worries, I know too well how easy it is to miss a PR. Thank you for the update!

PR #455, opened 2020-05-12
tileservices: add Cache-Status HTTP header to inform HIT or MISS on cached tiles

cunha17:
Once enabled by the config directive show_cache_status, the non-standard
HTTP header Cache-Status will provide information whether the response data
origin is cached data (HIT) or not (MISS).
Also, on each thousand requests made to the tile services, it will be logged the
ratio (%) of hits and misses.

### DIFF
--- a/mapproxy/cache/tile.py
+++ b/mapproxy/cache/tile.py
@@ -36,6 +36,9 @@ Tile caching (creation, caching and retrieval of tiles).
 """
 
 
+import logging
+log = logging.getLogger('mapproxy.config')
+
 from functools import partial
 from contextlib import contextmanager
 from mapproxy.grid import MetaGrid
@@ -64,6 +67,7 @@ class TileManager(object):
             bulk_meta_tiles=False,
             rescale_tiles=0,
             cache_rescaled_tiles=False,
+            show_cache_status=False
         ):
         self.grid = grid
         self.cache = cache
@@ -79,6 +83,10 @@ class TileManager(object):
         self.pre_store_filter = pre_store_filter or []
         self.concurrent_tile_creators = concurrent_tile_creators
         self.tile_creator_class = tile_creator_class or TileCreator
+
+        self.show_cache_status = show_cache_status
+        self.cache_hits = 0
+        self.cache_misses = 0
 
         self.rescale_tiles = rescale_tiles
         self.cache_rescaled_tiles = cache_rescaled_tiles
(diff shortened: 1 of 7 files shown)


mapproxy/cache/tile.py
```diff
@@ -162,10 +170,19 @@ def _load_tile_coords(self, tiles, dimensions=None, with_metadata=False,
...
+                    # missing or staled
+                    uncached_tiles.append(tile)
+                    tile.cache_hit = False if self.show_cache_status == True else None
+                    self.cache_misses = self.cache_misses + 1
+                else:
+                    tile.cache_hit = True if self.show_cache_status == True else None
+                    self.cache_hits = self.cache_hits + 1
+        cache_count = self.cache_hits + self.cache_misses
+        if (cache_count % 1000 == 0):
+            log.info("Tile caching statistics: HITS: %d (%0.1f%)\t MISSES: %d (%0.1f%)" %
```
weskamm, line 184: Seems that this is logged even with show_cache_status=False.
Not sure if this is intended, because this may "spam" the logs and cannot be disabled

4msheinbergon:
@weskamm Any idea why this PR wasn't merged? It sounds like a really good idea/feature. 
I'd be happy to help, if any is required.

weskamm:
Did you see my comment? I guess there is a check missing around the logging if logs should be produced at all.
Besides, tests and documentation for this new feature would be required. Feel free to jump in!

olt:
I'm not sure the logging feature is that useful, even if configurable. 
- It will reset on each restart of the process (which can happen regularly, depending on the deployment method). 
- You will get multiple logs of the same count when using multi-threaded deployment methods (race condition).
- You will get multiple logs of different counts when using multi-process deployment methods (globals are note shared).

Just add the headers and write a WSGI middleware that moves the stats into something like Prometheus.

sheinbergon:
@weskamm. Also, I think there's a bug in the original implementation (I've re-implemented myself and found the cache status header was not working until I applied a fix).

@olt I totaly agree with you. I don't really mind the logging/count feature, it should not be here. the `Cache-Status`  header is what I care about.

I'm willing to contribute a cleaner PR (header addition only) and documentation, If you'd like me to :-)

sheinbergon:
@weskamm. Also, I think there's a bug in the original implementation (I've re-implemented myself and found the cache status header was not working until I applied a fix).

@olt I totaly agree with you. I don't really mind the logging/count feature, it should not be here. the `Cache-Status`  header is what I care about.

I'm willing to contribute a cleaner PR (header addition only) and documentation, If you'd like me to :-)

tobwen:
@sheinbergon 
> I'm willing to contribute a cleaner PR (header addition only) and documentation, If you'd like me to :-)

I don't have anything to decide, but I'd be highly interested in it.

sheinbergon:
@weskamm @olt any comments regarding my suggestion?

sheinbergon:
@weskamm @olt ?

4msheinbergon:
@weskamm @olt any comments regarding my suggestion?

weskamm:
Feel free to make another pr with documentation and tests included. seems like a useful feature to me

weskamm:
ok will close this now as its superseeded

PR #456, opened 2020-05-12
config: add support for requesting vnd.jpeg-png mixed type to geoserver

cunha17:
Geoserver supports client requests for mixed image type that delivers
png or jpeg according to transparency and full coverage of tiles. This
can reduce effectively the size of the data sent to clients. The vendor
specific mime type is 'vnd.jpeg-png', and this avoids the usage of
image/png as a mixed type that, when the image is effectively jpeg at
geoserver side, will need another conversion back to jpeg at mapproxy side.


olt:
Thanks!

PR #461, opened 2020-05-24
Fix SyntaxWarning issues with Python 3.8.

sebastic:
A few SyntaxWarning issues are reported when installing the Debian packages with python3.8:

 * SyntaxWarning: 'bool' object is not callable; perhaps you missed a comma?
 * SyntaxWarning: "is" with a literal. Did you mean "=="?


olt:
Thanks!

PR #463, opened 2020-05-28
Add time-to-live / automatic tile expiry to sqlite cache backend

werrolf:
Changes in this pull add a new allowable configuration value `ttl` to the `sqlite`-type cache and modify the load_tile / load_tiles methods to no longer find tiles older than this `ttl`.  
This is an automatic expiry time in seconds (comparable to `redis`-type cache `default_ttl`). For continuity, the default is zero, and is interpreted to mean infinite time-to-live / unlimited tile reuse.

Ttl-based expiry produces rolling, demand-driven automatic tile renewal, reducing the scenarios that would otherwise demand seed tasks.


weskamm:
Thanks for this nice feature!

PR #465, opened 2020-07-02
Add detail to RequestErrors caused by MapBBOXError

walkermatt:
Enable users to more easily determine the cause of a `MapBBOXError` caused by a large request by providing details of the reason.

I spent a while a few weeks ago trying to determine the cause of service exceptions related to WMS requests that where either `num_tiles > max_tile_limit` or dimensions > `max_output_pixels`.


olt:
Thanks Walker.

Issue #467, opened 2020-07-13
`escape` no longer included in cgi module


stuart23:
The `escape` function is no longer in the cgi module as of Python 3.8.

To get around this, I am currently building containers with this:
`RUN echo "from html import escape" >> `python -c "import cgi; print(cgi.__file__)"``

Obviously this is not ideal, and the escape function should be imported from `html` instead. Only difference here is that the `quote` kwarg defaults to True in html, and False in cgi.

https://docs.python.org/3.7/library/cgi.html#cgi.escape
https://docs.python.org/3.8/library/html.html#html.escape

martijn-heil:
Duplicate of #462

PR #468, opened 2020-07-15
Use compat imports for `cgi.escape` and `cgi.parse_qsl` Fix #462

### LINKED ISSUE #462: Python 3.8 compatibility
## Context
On Ubuntu 20.4 with python 3.8 .. mapproxy 1.12 cannot work fine without some modifications

## Expected Behavior
It should simply work

## Actual Behavior
errors like this:
```
AttributeError: module 'cgi' has no attribute 'parse_qsl'
ImportError: cannot import name 'escape' from 'cgi' (/usr/lib/python3.8/cgi.py)
```

## Possible Fix
According to
https://stackoverflow.com/a/58871055
I've modified the cgi.escape to html.escape cgi.parse_qsl to urllib.parse.parse_qsl
It's working now

## Your Environment
* Version used: MapProxy==1.12.0 from apt python3-mapproxy package
* Environment name and version: Python 3.8.2
* Server type and version: server-develop with some testing configs
* Operating System and version: Ubuntu 20.04 LTS

martijn-heil:
As required for Python 3.8 or later
This should fix #462

As of yet still totally untested, but I think it should work.
Anyone is free to test, I'll try it myself as well, when I have the time, but that might be some time later.

### DIFF
--- a/mapproxy/util/ext/tempita/__init__.py
+++ b/mapproxy/util/ext/tempita/__init__.py
@@ -33,6 +33,7 @@ from __future__ import print_function
 import re
 import sys
 import cgi
+from mapproxy.compat.modules import escape
 import os
 import tokenize
 from io import StringIO, BytesIO
@@ -437,10 +438,10 @@ def html_quote(value, force=True):
     if not isinstance(value, basestring_):
         value = coerce_text(value)
     if sys.version >= "3" and isinstance(value, bytes):
-        value = cgi.escape(value.decode('latin1'), 1)
+        value = escape(value.decode('latin1'), 1)
         value = value.encode('latin1')
     else:
-        value = cgi.escape(value, 1)
+        value = escape(value, 1)
     if sys.version < "3":
         if is_unicode(value):
             value = value.encode('ascii', 'xmlcharrefreplace')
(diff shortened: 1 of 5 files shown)


martijn-heil:
There is a slight difference between the functionality of the old `cgi.escape` and `html.escape`, as mentioned by @stuart23 in #467 
> Only difference here is that the quote kwarg defaults to True in html, and False in cgi.

This happens to be the very reason that `cgi.escape` was deprecated in the first place: [according to the documentation](https://docs.python.org/3.7/library/cgi.html#cgi.escape)
> Deprecated since version 3.2: This function is unsafe because quote is false by default, and therefore deprecated. Use html.escape() instead.

So all in all it's probably better to indeed use `html.escape` without further specifying the quote argument, but if there are any specific cases in my changes where the original authors depended on this old behaviour we need to further specify the argument then.

mapproxy/util/ext/tempita/__init__.py
```diff
@@ -33,6 +33,7 @@ def foo(bar):
 import re
 import sys
 import cgi
+from mapproxy.compat.modules import escape
```
olt, line 36: Can you move this down a few lines.

mapproxy/util/ext/tempita/__init__.py
```diff
@@ -33,6 +33,7 @@ def foo(bar):
 import re
 import sys
 import cgi
```
olt, line 35: Can you remove the cgi import, or is it used elsewhere?

PR #471, opened 2020-09-28
URL Prefix configuration

stuart23:
Added an option to add a prefix so that the app can be hosted at non root locations (e.g. at a route defined by a load balancer).

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -1047,6 +1047,16 @@ Configuration options for the TMS/Tile service.
   was not changed.
 
 
+``prefix``
+"""""""""""""
+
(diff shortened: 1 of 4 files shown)


olt:
Thanks for your contribution.

You can solve this by setting "X-Script-Name": https://mapproxy.org/docs/nightly/deployment.html#http-proxy
That way the service will also work even when accessed directly (for debugging), or when you need it on multiple different URL prefixes.

stuart23:
Thanks @olt. At the end of the day, it was easiest to inject that header using a middleware as injecting a header on one route with the NGINX k8s ingress controller isn't as straightforward.

PR #482, opened 2020-12-17
Add support for HTTP cookies

julienr:
This is based on initial work by @awaterme in #345, which was closed due to inactivity. I'm happy to take over and try to get this merged as we have the same need as Andreas.

This is the same code as #345, but I've addressed the comments about documentation and squashed all the commits. If you prefer that I leave Andreas as the author of the squashed commit, I can also do that (my intent is not to steal any credits, so please tell me what's best). Or @awaterme, if you actually want to continue working on your PR, happy to help you then.

This enables for sources that require cookies to manage session (for example rate-limiting clients based on number of sessions opened).

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -1024,6 +1024,15 @@ Add additional HTTP headers to all requests to your sources.
 
 Sets the ``Access-control-allow-origin`` header to HTTP responses for `Cross-origin resource sharing <http://en.wikipedia.org/wiki/Cross-origin_resource_sharing>`_. This header is required for WebGL or Canvas web clients. Defaults to `*`. Leave empty to disable the header. This option is only available in `globals`.
 
+``manage_cookies``
+^^^^^^^^^^^^^^^^^^
+
+.. versionadded:: 1.14.0
+
+Enables MapProxy cookie management for HTTP sources. When enabled MapProxy will accept and store server cookies. Accepted cookies will be passed
+back to the source on subsequent requests. Usefull for sources which require to maintain an HTTP session to work efficiently, maybe in combination
+with basic authentication. Depending on your deployment MapProxy will still start multiple sessions (e.g. one per MapProxy process).
+Cookie handling is based on Python `CookieJar <https://docs.python.org/3/library/http.cookiejar.html>`_. Disabled by default.
 
 ``hide_error_details``
 ^^^^^^^^^^^^^^^^^^^^^^
--- a/doc/configuration_examples.rst
+++ b/doc/configuration_examples.rst
@@ -708,6 +708,8 @@ You can disable the certificate verification if you you don't need it.
       url: https://username:mypassword@example.org/service?
       layers: securelayer
 
+.. note:: If the source requires session handling through cookies, have a look at the ``manage_cookies`` configuration option.
+
 .. _http_proxy:
 
 Access sources through HTTP proxy
@@ -734,10 +736,27 @@ You can also set this in your :ref:`server script <server_script>`::
 
 Add a username and password to the URL if your HTTP proxy requires authentication. For example ``http://username:password@example.com:3128``.
 
+.. note:: If the source requires session handling through cookies, have a look at the ``manage_cookies`` configuration option.
+
 You can use the ``no_proxy`` environment variable if you need to bypass the proxy for some hosts::
 
   $ export no_proxy="localhost,127.0.0.1,196.168.1.99"
 
+
+Cookie Management
+=================
+
+MapProxy can handle server cookies of HTTP sources, like browsers do. That is, MapProxy accepts cookies and passes them back
+on subsequent calls. This is useful for sources that use cookie for session management or rate-limiting for example::
+
+  sources:
+    wms_with_session_management:
+      type: wms
+      http:
+        manage_cookies: True
+      req:
+        url: http://example.org/service?
+        layers: layer0
 
 .. _paster_urlmap:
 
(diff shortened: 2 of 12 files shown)


doc/configuration.rst
```diff
@@ -1024,6 +1024,15 @@ Add additional HTTP headers to all requests to your sources.
 
 Sets the ``Access-control-allow-origin`` header to HTTP responses for `Cross-origin resource sharing <http://en.wikipedia.org/wiki/Cross-origin_resource_sharing>`_. This header is required for WebGL or Canvas web clients. Defaults to `*`. Leave empty to disable the header. This option is only available in `globals`.
 
+``manage_cookies``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. versionadded:: 1.12.0
```
weskamm, line 1030: should read 1.14.0 (as the current latest version is 1.13)

awaterme:
Hi Julien,
I'm very glad somebody picked up the issue again. Please go ahead  and don't worry about my credits.
Best regards,
Andreas

mapproxy/config_template/base_config/full_example.yaml
```diff
@@ -425,10 +425,27 @@ sources:
     req:
       # username and password are extracted from the URL and do not show
       # up in log files
+      # ATTENTION: see https://github.com/mapproxy/mapproxy/issues/182
```
weskamm, line 428: please remove this line and move the information to the docs in the appropriate section

mapproxy/config_template/base_config/full_example.yaml
```diff
@@ -425,10 +425,27 @@ sources:
...
 
+  # WMS source that requires authentication and session management
+  # through HTTP cookies
+  session_source:
+    type: wms
+    http:
+      # Accept session cookies and forward on subsequent requests
+      manage_cookies: true
+      # Use basic auth header directly
+      # see https://github.com/mapproxy/mapproxy/issues/182
```
weskamm, line 441: same as above, please remove this line

weskamm:
Thanks for picking this up again.
Just found 2 minor issues, which you may want to adress. Else this is looking good!

doc/configuration.rst
```diff
@@ -1024,6 +1024,15 @@ Add additional HTTP headers to all requests to your sources.
 
 Sets the ``Access-control-allow-origin`` header to HTTP responses for `Cross-origin resource sharing <http://en.wikipedia.org/wiki/Cross-origin_resource_sharing>`_. This header is required for WebGL or Canvas web clients. Defaults to `*`. Leave empty to disable the header. This option is only available in `globals`.
 
+``manage_cookies``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
olt, line 1028: .rst requires that the ^^^ line length matches the title.

doc/configuration_examples.rst
```diff
@@ -734,11 +736,28 @@ You can also set this in your :ref:`server script <server_script>`::
...
 
+.. note:: If the source requires session handling through cookies, have a look at the ``manage_cookies`` configuration option.
+
 You can use the ``no_proxy`` environment variable if you need to bypass the proxy for some hosts::
 
   $ export no_proxy="localhost,127.0.0.1,196.168.1.99"
 
 
+Cookie Management
+==================
```
olt, line 747: Same here.

olt (commented):
A feature like this should have at least a small system test to assure that it does work now and more importantly in the future.

julienr:
Thanks for the reviews. I've added a test for the http client behavior of Set-Cookie. I'll just add another one in `test_conf_loader` checking that the config parameter is correctly passed from config file to http client.

doc/configuration.rst
julienr: Good catch - I've fixed it

mapproxy/config_template/base_config/full_example.yaml
julienr: Removed

mapproxy/config_template/base_config/full_example.yaml
julienr: Removed

julienr:
> A feature like this should have at least a small system test to assure that it does work now and more importantly in the future.

I've added a system test to `test_wms` and two unit test to `test_client.py` (to check that enabling/disabling the functionality has the correct behavior). 

Is it the kind of tests you were expecting or should I do it differently ?

doc/configuration.rst
julienr: Good catch - fixed

doc/configuration_examples.rst
julienr: Good catch - fixed

weskamm (commented):
The tests are failing, but it looks like its easy to fix.
I think the idea of testing on and off behaviour is good.

I am a little bit confused about the structure of your tests, as it looks somewhat different to all the other tests. But this is maybe just a personal taste.
So if the tests pass again, i think this could be merged. Thanks!

julienr:
@weskamm I've fixed the tests - sorry for pushing something broken.

As far as test structure go, are you referring to the `assert_cookie` / `assert_no_cookie` functions ? I've done it that way (passing a helper to the mock request setup) because that's the only way I found to access the request from the mock server side. But if there is a better way, I'd be happy to do it.

weskamm:
I was hoping that someone else could answer to that question, but i think its ok. Thanks!

weskamm (approved):

PR #484, opened 2021-01-06
Fix SQL Expression parsing errors with GDAL 2.2+

smomapz:
Since GDAL 2.2 layer names loaded from geojson datasources using OGR are no longer always 'OGRGeoJSON'. Since they can contain characters that can mess with the SQL string to apply the where-clause they need to be quoted.


walkermatt:
It would be fantastic if we had tests to cover this behaviour, would you be able to look into this?

Current tests are here: [mapproxy/test/unit/test_ogr_reader.py](https://github.com/mapproxy/mapproxy/blob/97cf9763b3a6d16e647d8e8c5dce0f94ce516044/mapproxy/test/unit/test_ogr_reader.py). Currently the test loads a Shapefile, a new test could load a GeoJSON with a filename that causes the current implementation to fail?

smomapz:
I added a geojson testfile and updated the unit test. I don't know much about unit tests in python though, so please let me know if this is done well enough.

Also this could be tested and reproduced a lot easier with just renaming the existing shapefile to contain a character that is not valid in SQL string like polygons-germany.shp. This would keep the tests and the test data smaller but would not test the OGRReader with geojson files. Let me know if this approach suits you better.

smomapz:
Anything still missing for this to get released?

smomapz:
@walkermatt Bump for visibility. We really need this merged. Is anything missing still?

walkermatt:
@smomapz if you could update the GeoJSON test so that the filename causes the test to fail prior to your quoting change I think this is good to go.

smomapz:
@walkermatt The test is already failing prior to my changes. The filename is not relevant when the geojson file provides a layer name. Here with "polygons-germany" the hyphen is causing the issue in the old implementation.

walkermatt:
@smomapz thanks, I was able to reproduce the issue and confirm the fix this afternoon. I've squashed your commits and merged :+1:

Issue #478, opened 2020-11-17
Introduce linting


weskamm:
It would be nice to have linting (pylint?) included in mapproxy.
PEP-8 rules should be followed (https://www.python.org/dev/peps/pep-0008/)
Enforcing of linting through travis for PRs would be nice too.

Of course this will lead to a bigger initial Pull Request...

simonseyock:
flake8 is an alternative. Maybe https://github.com/hhatto/autopep8 can be used?

julienr:
I'd be happy to have a look into this if nobody else is on it yet. I've some past experience at work with setting up flake8 on CI, so I could give this a shot.

> Of course this will lead to a bigger initial Pull Request...

One option to avoid thousand lines PR would be to start with a flake config file that disable most rules and then start enabling 1 by 1 - breaking each one into its own commit/PR.

weskamm:
I like the idea of splitting it up. Feel free to go ahead!

julienr:
I started experimenting with this in #486

olt:
You should bring this to the PSC as it affects project infrastructure and developing policies.
https://github.com/mapproxy/mapproxy/wiki/PSC-Guidelines

PR #486, opened 2021-01-11
Add flake8 for linting on CI

julienr:
This is a first step towards introducing linting (#478) using flake8. The strategy is to start by ignoring all errors and un-ignore them step by step to break this into multiple commits.

For now, I'm using this PR to make sure it works with the CI. I will undraft once I get it to work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,4 +60,5 @@ install:
     - "pip freeze"
 
 script:
+    - flake8 mapproxy
     - pytest mapproxy
--- a/setup.cfg
+++ b/setup.cfg
@@ -4,6 +4,69 @@ doctest_optionflags= NORMALIZE_WHITESPACE IGNORE_EXCEPTION_DETAIL ALLOW_UNICODE 
 norecursedirs = fixture .git .tox
 testpaths = mapproxy
 
+[flake8]
+ignore = E111
+         E117
+         E121
+         E122
+         E123
+         E124
+         E125
+         E126
+         E127
+         E128
+         E129
+         E131
+         E201
+         E202
+         E203
+         E221
+         E222
+         E223
+         E225
+         E226
+         E228
+         E231
+         E241
+         E251
+         E261
+         E262
+         E265
+         E266
+         E271
+         E272
+         E301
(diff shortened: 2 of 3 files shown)


setup.cfg
```diff
@@ -4,6 +4,9 @@ doctest_optionflags= NORMALIZE_WHITESPACE IGNORE_EXCEPTION_DETAIL ALLOW_UNICODE
 norecursedirs = fixture .git .tox
 testpaths = mapproxy
 
+[flake8]
+ignore = E117,E121,E122,E123,E124,E125,E126,E127,E128,E129,E131,E201,E202,E203,E221,E222,E223,E225,E226,E228,E231,E241,E251,E261,E262,E265,E266,E271,E272,E301,E302,E303,E305,E306,E402,E501,E502,E701,E702,E703,E711,E712,E713,E721,E722,E731,E741,F401,F523,F812,F821,F841,F901,W503,W504,W191,W291,W292,W293,W391
```
olt, line 8: It would be more helpful (in the long term) if you put each code on a separate line and also add a comment for each code. This will make future changes easier (diffs).

Maybe you find a list of all codes that you can copy&paste?

julienr: I've updated the PR to have one code per line. I indeed got the list of code by running `flake8 --statistics` and picking all the codes that yield at least one error.

I'll send an email to the mailing list to get PSC approval before I start looking into fixing each code one by one.

.travis.yml
```diff
@@ -60,4 +60,5 @@ install:
     - "pip freeze"
 
 script:
+    - flake8 mapproxy
```
julienr, line 63: To check: not sure if it would be desirable to make the build fail early when there are linting errors. If we want that, I just need to check how to do that with travis. I feel the easy way would be `flake8 mapproxy && pytest mapproxy`, but maybe there is a more travis-elegant way.

weskamm: dont rely on travis, it may get replaced soon, see https://github.com/mapproxy/mapproxy/issues/487

weskamm:

Issue #495, opened 2021-03-06
[has fix] issues with capabilities documents behind reverse proxy // extended


tobwen:
## Context
I'm running  _MapProxy_ behind a reverse proxy under its own path (eg. `/mapproxy`), but some issue appear.

## Expected Behavior
HTML-flavour _GetCapabilities_ documents should be shown.

## Actual Behavior
_GetCapabilities_ (HTML) and _Layer Capabilities_ raise an `internal error`, since the local templates can't be found (due to the missing `SCRIPT_NAME` variable).

## Possible Fix
#553

## Steps to Reproduce
1. Setup _NGINX_ (or _Apache_) as a reverse-proxy for _MapProxy_ (taken from docs):
```nginx
server {
  server_name example.org;
  location /mapproxy {
    proxy_pass http://127.0.0.1:8000;
    proxy_set_header Host $http_host;
    proxy_set_header X-Script-Name /mapproxy;
    proxy_set_header SCRIPT_NAME /mapproxy; # testing
  }
}
```
2. Start _MapProxy's_ dev-server: `mapproxy-util serve-develop mapproxy.yaml -b 0.0.0.0:8000`
3. Visit the demo website with your browser: `http://127.0.0.1:80/mapproxy/demo`
4. Click on the HTML-flavor _Capabilities document_, it's failing with _internal error_. Also, the _Layer Capabilities_ are failing.
```

## Your Environment
* Version used: https://github.com/mapproxy/mapproxy/commit/3a5d3844faa90065c50110edccaa53c34644c998
* Environment name and version: Python 3.7.3 (same issue with 3.8 and 3.9) 
* Server type and version: bare metal
* Operating System and version: latest Debian Buster, AMD64

tobwen:
Passing `SCRIPT_NAME=/mapproxy` to an WSGI server like _gunicorn_ makes the first part work (the documents still point to the internal addresses). Seems like `SCRIPT_NAME` needs to be present when the application starts?

tobwen:
# updated and extended
The user can now choose to view the capabilities documents either for the internal address (e.g. 10.0.13.37) or the external address (e.g. example.org). This makes debugging easier when the service is running behind a reverse proxy and throws an "internal error" as before.

![image](https://user-images.githubusercontent.com/1864057/146301534-eb3f8505-c584-4756-929f-9ffcb9909e7b.png)

PR #497, opened 2021-03-06
fix capability documents when behind reverse proxy

tobwen:
Fixes https://github.com/mapproxy/mapproxy/issues/495

### DIFF
--- a/mapproxy/request/base.py
+++ b/mapproxy/request/base.py
@@ -196,7 +196,7 @@ class Request(object):
         if script_name:
             del environ['HTTP_X_SCRIPT_NAME']
             environ['SCRIPT_NAME'] = script_name
-            path_info = environ['PATH_INFO']
+            path_info = environ.get('PATH_INFO', '')
             if path_info.startswith(script_name):
                 environ['PATH_INFO'] = path_info[len(script_name):]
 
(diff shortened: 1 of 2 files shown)


mapproxy/request/base.py
```diff
@@ -196,7 +196,7 @@ def __init__(self, environ):
         if script_name:
             del environ['HTTP_X_SCRIPT_NAME']
             environ['SCRIPT_NAME'] = script_name
-            path_info = environ['PATH_INFO']
+            path_info = environ.get('PATH_INFO', '')
```
tobwen, line 199: Maybe not needed, but I had issues with it.

mapproxy/request/base.py
```diff
@@ -279,7 +279,7 @@ def server_script_url(self):
         return self.script_url.replace(
             self.host_url.rstrip('/'),
             self.server_url.rstrip('/')
-        )
+        )[:-len(self.environ.get('SCRIPT_NAME', ''))]
```
tobwen, line 282: When behind a reverse proxy, this gets rid of `SCRIPT_NAME`, which doesn't exist on the backend.

tobwen:
Closed due to massive refactoring.

PR #514, opened 2021-05-18
Fix caching

mki-c2c:
Cache handling in the Navigator is confusing since the cache headers are not correctly recognized if not using HTTP1.1

This PR fixes cache problems in the Navigator and fixes another roundoff error observed in caching with NOT-MODIFIED-SINCE header

### DIFF
--- a/mapproxy/util/ext/serving.py
+++ b/mapproxy/util/ext/serving.py
@@ -123,7 +123,6 @@ class WSGIRequestHandler(BaseHTTPRequestHandler, object):
 
         def write(data):
             assert headers_set, 'write() before start_response'
-            import pdb; pdb.set_trace()
             if not headers_sent:
                 if self.request_version == "HTTP/1.0":
                     headers_sent[:] = headers_set
--- a/mapproxy/util/ext/serving.py
+++ b/mapproxy/util/ext/serving.py
@@ -62,6 +62,8 @@ def _log(type, message, *args):
 class WSGIRequestHandler(BaseHTTPRequestHandler, object):
     """A request handler that implements WSGI dispatching."""
 
+    protocol_version = 'HTTP/1.1'
+
     @property
     def server_version(self):
         return 'MapProxy/' + mapproxy.version.__version__ + ' (Werkzeug based)'
--- a/mapproxy/util/ext/serving.py
+++ b/mapproxy/util/ext/serving.py
@@ -127,15 +127,15 @@ class WSGIRequestHandler(BaseHTTPRequestHandler, object):
                 if self.request_version == "HTTP/1.0":
                     headers_sent[:] = headers_set
                     status = headers_set[0]
-                    # remove non-HTTP1.0 headers
+                    # only keep HTTP/1.0 headers
                     response_headers = (
                         header_tuple
                         for header_tuple in headers_set[1]
                         if header_tuple[0].lower() in [
-                                "date", "pragma",
-                                "location", "server", "www-authenticate",
-                                "allow", "content-Encoding", "content-length",
-                                "content-type", "expires", "last-modified"
+                            "date", "pragma",
+                            "location", "server", "www-authenticate",
+                            "allow", "content-encoding", "content-length",
+                            "content-type", "expires", "last-modified"
                         ]
                     )
                 else:
@@ -252,7 +252,7 @@ class WSGIRequestHandler(BaseHTTPRequestHandler, object):
         self.log_request(code)
         if message is None:
             message = code in self.responses and self.responses[code][0] or ''
-        if self.request_version == 'HTTP/0.9':
+        if self.request_version != 'HTTP/0.9':
             if self.request_version == 'HTTP/1.0':
                 hdr = "%s %d %s\r\n" % (self.request_version, code, message)
             else:
--- a/mapproxy/util/ext/serving.py
+++ b/mapproxy/util/ext/serving.py
@@ -22,9 +22,11 @@ except ImportError:
     import _thread as thread
 
 try:
+    # python 2
     from SocketServer import ThreadingMixIn
     from BaseHTTPServer import HTTPServer, BaseHTTPRequestHandler
 except ImportError:
+    # python 3
     from socketserver import ThreadingMixIn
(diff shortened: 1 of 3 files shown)


ramunasd:
Looks good but tests are missing :)

mki-c2c:
alright, I'll look into the tests

mki-c2c:
I updated the tests: introduced a non integer timestamp in `_update_timestamp` fixture

I have non integer timestamps on the local file system and in this case, `wsgiref.handlers.format_date_time` invoked by `mapproxy.util.time.format_httpdate` cuts off the fractional part of the date seconds. This makes the test `test_if_modified_since` fail.
the test case fails without the integer cast introduced in 81b3d31

mki-c2c:
I could not easily test the http1.1 compliance.
I see one issue though. The cache handling with etag and last-modified headers is not compliant with HTTP1.0, so browsers were confused when receiving an HTTP1.0 response and may have ignored the etag.

With the currect modification, the server replies with an HTTP1.1 response, even if the client requests HTTP1.0. However, HTTP1.1 elements cannot easily be removed in the response.
I would just leave it like this, it works well on my (recent) browsers chrome, firefox. No idea which should be the behaviour following an explicit HTTP1.0 request ?

weskamm:
I agree that this a needed fix, as most requests will use HTTP 1.1, but why cant we just remove invalid headers for http 1.0 responses when we detect such requests? I think removing uncompliant headers (like transfer-encoding, etc) should be sufficient?

mki-c2c:
thanks a lot for your review, I'll have a look into this at the beginning of next week.

mki-c2c:
Hello, I looked a bit into HTTP1.0 compliance.
It starts getting really old, so I'm not sure if it deserves much efforts.
Even google returns a HTTP1.0 response with unsupported headers.

however, I tried to implement return of a clean HTTP1.0 response by removing unsupported headers. This works, and I checked it manually, but there are no CI tests on the complete WGSI server, only on a higher level, which does not care about HTTP version.
I am not sure about the cost of the HTTP version check at every request. Probably there is no big impact, but I cannot assess the mapproxy server performance very well.

Actually, I would prefer to either ignore HTTP1.0 requests and answer with an HTTP1.1 response, and let the client cope with that. Or maybe just do like google.com, return a response with HTTP1.0 status line, but with all the same headers as in HTTP1.1

What do you think about merging the PR with or without the latest commit https://github.com/mapproxy/mapproxy/pull/514/commits/69160c1933e978ef4c6cac12a4069a3e9b8106ad ?

mapproxy/util/ext/serving.py
```diff
@@ -119,8 +123,24 @@ def run_wsgi(self):
 
         def write(data):
             assert headers_set, 'write() before start_response'
+            import pdb; pdb.set_trace()
```
weskamm, line 126: this should be removed

mapproxy/util/ext/serving.py
```diff
@@ -119,8 +123,24 @@ def run_wsgi(self):
...
             if not headers_sent:
-                status, response_headers = headers_sent[:] = headers_set
+                if self.request_version == "HTTP/1.0":
+                    headers_sent[:] = headers_set
+                    status = headers_set[0]
+                    # remove non-HTTP1.0 headers
+                    response_headers = (
```
weskamm, line 132: i cannot see any removal here, i think you are doing the opposite and keep all unwanted headers

mapproxy/util/ext/serving.py
```diff
@@ -119,8 +123,24 @@ def run_wsgi(self):
...
+                        header_tuple
+                        for header_tuple in headers_set[1]
+                        if header_tuple[0].lower() in [
```
weskamm, line 135: i think we need a `not` in here

mapproxy/util/ext/serving.py
mki-c2c: ouch, yes of course, that slipped through and there is not CI test of the http server to catch this.
Thanks

mapproxy/util/ext/serving.py
```diff
@@ -119,8 +123,24 @@ def run_wsgi(self):
...
+                                "date", "pragma",
+                                "location", "server", "www-authenticate",
+                                "allow", "content-Encoding", "content-length",
```
weskamm, line 138: please make everything lowercase in here ("content-Encoding") or it will not match, as you are comparing with lowerCase values

mapproxy/util/ext/serving.py
```diff
@@ -233,8 +253,11 @@ def send_response(self, code, message=None):
         self.log_request(code)
         if message is None:
             message = code in self.responses and self.responses[code][0] or ''
-        if self.request_version != 'HTTP/0.9':
-            hdr = "%s %d %s\r\n" % (self.protocol_version, code, message)
+        if self.request_version == 'HTTP/0.9':
```
weskamm, line 255: this `if` does not make sense, i think we need this to read `!=` instead of `==`

mapproxy/util/ext/serving.py
mki-c2c: OK, I'll update the comment. The list contains all http 1.0 headers I found, so the logic is correct, but is not in line with the comment

mapproxy/util/ext/serving.py
mki-c2c: yes, thanks

mapproxy/util/ext/serving.py
mki-c2c: yes, I had reorganized the code, and left this typo

weskamm:
In general i think this is going into the right direction, but i am unsure and maybe not the right one to answer the question, if setting the `protocol_version` on a fixed value like it is done now is a good idea. I guess before this change, mapproxy always answered with version 1.0 as its the default from the inherited class. But i am unsure how relevant the setting is, as the surrounding webserver / wsgi server like nginx, gunicorn etc. may upgrade the protocol anyways?!

Maybe it would be better to return a response using the same protocol the request has been made (self.request_version), but like i said this maybe answered better by another one more involved in this...

weskamm:

PR #515, opened 2021-05-18
mapproxy seed - CL Option skip uncached

mki-c2c:
This feature covers a use case where mapproxy-seed is run to update stale tiles in the cache.
All code is already present, the feature is just made available via command-line options.


mki-c2c:
tests missing, so the PR remains draft for the time being.

weskamm (approved):

PR #518, opened 2021-06-02
Refresh while serving

mki-c2c:
This is an enhancement, so that MapProxy can be configured to refresh tiles from source while serving.

The expiration period of a given cache can be configured as in the seeding configuration:

- absoute time
- relative time
- file reference time


mki-c2c:
Hello @ramunasd, thanks for your thumbs up on the PR.
I have written to the community on the mailing list in order to get some feedback on how to integrate the added features of PR #515 and #518 into the main mapproxy branch, however I got no response.
What's the procedure and who should review the modifications before merging them ?
Would it be possible to be involved in the 2021 release ?

ramunasd:
Correct me if I'm wrong but this does not refresh tiles, just serves old ones on SourceError?

mapproxy/cache/tile.py
```diff
@@ -76,6 +78,7 @@ def __init__(self, grid, cache, sources, format, locker, image_opts=None, reques
         self.sources = sources
         self.minimize_meta_requests = minimize_meta_requests
         self._expire_timestamp = None
+        self._refresh_before = {}
```
ramunasd, line 81: how this var gets populated from config?

mapproxy/cache/tile.py
```diff
@@ -362,17 +376,24 @@ def _create_single_tile(self, tile):
...
-                    # call as_buffer to force conversion into cache format
-                    source.as_buffer(self.tile_mgr.image_opts)
-                source.image_opts = self.tile_mgr.image_opts
-                tile.source = source
-                tile.cacheable = source.cacheable
-                tile = self.tile_mgr.apply_tile_filter(tile)
-                if source.cacheable:
-                    self.cache.store_tile(tile)
+                try:
+                    source = self._query_sources(query)
```
ramunasd, line 380: perhaps it makes sense to try-catch only this one statement? everything else does not throw such exception

ramunasd (changes requested):

mki-c2c:
Up to now, mapproxy-serve does neither of these:
- If the tile is found in cache, it is served, whatever its age.
- If the tile is not found in cache, it is requested from source.
  - if the source is available, it is served and cached:
  - if the source is not available, either of following happens:
    - if an error handler applies (on_error), an error tile is generated (and optionally cached)
    - if there is no error handler, you get an http error

So the server has currently no notion of stale tiles.

What is new in this PR:
- If tile is found in cache, but expired according to the new parameter `refresh_before`:
   - a new tile is requested.
      - if source available: new tile is served and cache refreshed
      - if source unavailable: generation of error tile or http error as above

I am currently working on the error behaviour, to be able to optionally serve a stale tile, if tile is expired and source is unavailable, see https://github.com/camptocamp/mapproxy/pull/1

mapproxy/cache/tile.py
mki-c2c: Yes, that's a good idea, it also makes sense with further work on https://github.com/camptocamp/mapproxy/pull/1

mapproxy/cache/tile.py
mki-c2c: It is set in [loader.py#L1578](https://github.com/mapproxy/mapproxy/blob/65d05bf34ba4a0928043b0dfc1963e40e3ae332a/mapproxy/config/loader.py#L1578)
This is not very pretty, but is a similar fashion as for mapproxy-seed which also externally sets this attribute af the tile manager:
- in seed - refresh: [seeder.py#L512](https://github.com/camptocamp/mapproxy/blob/refresh_while_serving/mapproxy/seed/seeder.py#L512)
- in seed- cleanup: [cleanup.py#L116](https://github.com/camptocamp/mapproxy/blob/refresh_while_serving/mapproxy/seed/cleanup.py#L116)

mapproxy/cache/tile.py
mki-c2c: done -> https://github.com/mapproxy/mapproxy/pull/518/commits/20db92c1084b9d115f5c863c65faf0fdefb19149

ramunasd (approved):

ramunasd:
Just one thought came to my head :) Can we set response headers to something 'max-age:0,must-revalidate' if we are serving stale tile? @mki-c2c

mki-c2c:
Well, it sounded like a good idea at first, but the question has been twisting my mind for a while now.
I think the service provider (mapproxy operator) chooses to authorize use of the stale tile for some time as it authorizes use of the valid tile for some time. The `expires_hours` will be chosen accordingly.
Maybe it is useful to configure a different value for expires_hours in the case of stale tiles.

I am somehow afraid of forcing no-cache for stale tiles, since the server will then receive a big amount of requests:
it breaks the browser caching strategies, since the browser has to re-request the stale tile every time.

Tricky questions, I have started a reply 3 times and each time I realized a new subtle detail.

mki-c2c:
Hello @weskamm,
I have been working on some additional features for mapproxy, and I would like to discuss the possibilities to merge the features into an upcoming release.
I have not managed to get any feedback on the release schedule via the mailing list or the discussion on pull requests. Since you have just prepared two bugfix releases, I wonder if you have any further information about the roadmap for a 2021 feature release and how to participate in the next release.

weskamm:
Hi @mki-c2c,
i already noticed your PRs.
There currently is no roadmap for upcoming releases, they happen when time and features allows or urgency is high.
In other words: I cannot tell you if or when your PRs will get merged and released, but as far as i can see, most of them are looking good and have good chances, when tested and reviewed, to be included in the next release...

ramunasd:
@weskamm I will add this to next milestone 1.14
Let's organize things better here :)

mapproxy/cache/tile.py
```diff
@@ -318,6 +326,12 @@ def is_cached(self, tile):
         """
         return self.tile_mgr.is_cached(tile)
 
+    def is_stale(self, tile):
+        """
+        Return True if the tile is cached.
```
weskamm, line 331: Should better read `Return True if tile exists _and_ is expired.` as the original method

weskamm (commented):
Just added a minor note, else this look like a nice addition, thanks!

mki-c2c:
took into account your remark on the comment, thanks.

PR #521, opened 2021-06-09
Fix TypeError when writing to stdout

neon-ninja:
This PR fixes the following TypeError when running the command `mapproxy-util autoconfig     --capabilities https://api-proxy.auckland-cer.cloud.edu.au/cgi-bin/mapserv`:

```python
  File "/usr/local/bin/mapproxy-util", line 11, in <module>
    load_entry_point('MapProxy', 'console_scripts', 'mapproxy-util')()
  File "/home/ubuntu/mapproxy/mapproxy/script/util.py", line 387, in main
    commands[command]['func'](args)
  File "/home/ubuntu/mapproxy/mapproxy/script/conf/app.py", line 185, in config_command
    write_header(f, options.capabilities)
  File "/home/ubuntu/mapproxy/mapproxy/script/conf/app.py", line 59, in write_header
    print('# MapProxy configuration automatically generated from:', file=f)
  File "/usr/lib/python3.8/codecs.py", line 378, in write
    self.stream.write(data)
TypeError: write() argument must be str, not bytes
```

See https://stackoverflow.com/a/44447780 for an explanation

### DIFF
--- a/mapproxy/script/conf/app.py
+++ b/mapproxy/script/conf/app.py
@@ -71,7 +71,7 @@ def write_header(f, capabilities):
 @contextmanager
 def file_or_stdout(name):
     if name == '-':
-        yield codecs.getwriter('utf-8')(sys.stdout)
+        yield codecs.getwriter('utf-8')(sys.stdout.buffer)
     else:
         with open(name, 'wb') as f:
             yield codecs.getwriter('utf-8')(f)
--- a/mapproxy/script/conf/app.py
+++ b/mapproxy/script/conf/app.py
@@ -71,7 +71,10 @@ def write_header(f, capabilities):
 @contextmanager
 def file_or_stdout(name):
     if name == '-':
-        yield codecs.getwriter('utf-8')(sys.stdout.buffer)
+        if hasattr(sys.stdout, 'buffer'):
+            yield codecs.getwriter('utf-8')(sys.stdout.buffer)
+        else:
+            yield codecs.getwriter('utf-8')(sys.stdout)
     else:
         with open(name, 'wb') as f:
             yield codecs.getwriter('utf-8')(f)


mapproxy/script/conf/app.py
```diff
@@ -71,7 +71,7 @@ def write_header(f, capabilities):
 @contextmanager
 def file_or_stdout(name):
     if name == '-':
-        yield codecs.getwriter('utf-8')(sys.stdout)
+        yield codecs.getwriter('utf-8')(sys.stdout.buffer)
```
mki-c2c, line 74: Caution, the integration tests failed on your modifications.
I think your line of code should be:
```
yield codecs.getwriter('utf-8')(sys.stdout.decode())
```

background: in python2 there was little distinction between byte streams and string. unicode was treated separately.
In python3, you go from bytes to str (unicode) via decode() and from str to bytes via encode()
https://stackoverflow.com/questions/14472650/python-3-encode-decode-vs-bytes-str

mki-c2c (changes requested):

mapproxy/script/conf/app.py
neon-ninja: That doesn't make any sense. We're writing to stdout, not reading from it.

```python
>>> sys.stdout.decode
  File "<stdin>", line 1, in <module>
AttributeError: '_io.TextIOWrapper' object has no attribute 'decode'
```

mki-c2c: yes, you're completely right, I didn't read the error message thoroughly enough. So maybe there is something wrong with the CI test, the fixtures, should have a look into that.
Probably your fix is correct, since you're testing it on a real life example. However something does not work in the automatic tests.

mki-c2c: There seems to be some trouble with using stdout in CI tests, for example a special case is made here:
https://github.com/mapproxy/mapproxy/blob/89217d05a2903c98a0e4bf7f91ac5e2dab5d94b0/mapproxy/script/conf/app.py#L48

I think your solution works fine on python3, but in python2 sys.stdout does not have a buffer attribute. It would be better to find a solution that works for both python2 and 3.

neon-ninja: https://github.com/mapproxy/mapproxy/pull/521/commits/99a8a1cb847112863cc685757057d800fd3d6d28 should fix this problem both in the tests and in Python2

mki-c2c: great! looks like a good hack.
Strange that CI has not started on your push.

neon-ninja: Take a look at https://travis-ci.org/github/mapproxy/mapproxy/requests - it says "Since June 15th, 2021, the building on travis-ci.org is ceased. Please use travis-ci.com from now on.".  Travis got the request but it rejected the job as Travis is in "Read-only mode"

weskamm:
Thanks for contribution.
Closing this one in favor of https://github.com/mapproxy/mapproxy/pull/544

PR #523, opened 2021-06-21
Allow reuse of Mapnik objects if enabled

Nakaner:
When Mapproxy uses Mapnik without using Renderd (by the way, that's not documented), a new Mapnik map object is initialised for any new rendering request. This harms performance – at least with some map styles – because Mapnik runs all the SQL queries of the layers of the style with an appended `LIMIT 0` to get the types of the returned columns. This pull request allows to reuse map objects.

For example, rendering a map style similar to OSM Carto requires 20 to 30 seconds without, and 6 to 8 seconds with reused Mapnik map objects. Without this patch, Mapproxy performance significantly slower with Mapnik sources compared to a Tirex+mod_tile setup.

Remarks: Multithreading is not supported. If a WSGI worker daemon is restarted, Mapnik is initialised when the first Mapnik rendering is requested.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -1063,6 +1063,18 @@ The complete path to the ``mapserv`` executable. Required if you use the ``mapse
 Path where the Mapserver should be executed from. It should be the directory where any relative paths in your mapfile are based on. Defaults to the directory of ``binary``.
 
 
+``mapnik``
+""""""""""
+
+Options for the :ref:`Mapnik source<mapnik_label>`.
+
+``reuse_map_objects``
+^^^^^^^^^^^^^^^^^^^^^
+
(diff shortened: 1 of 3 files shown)


mki-c2c (commented):
looks good. Could there be any protection to avoid having a config conflict between multithread and reuse mapnik ?

ArneBab:
Here’s also a larger (and therefore more risky) pull-request that adds multithreaded operation on mapnik. Maybe it could be useful for you: https://github.com/mapproxy/mapproxy/pull/447/

It uses precise caching and per-thread re-use of mapnik instances.

weskamm:
superseeded by https://github.com/mapproxy/mapproxy/pull/447

Issue #527, opened 2021-07-14
Remediation for #526 breaks demo-service


FrankyBoy:
Stacktrace from log:

```
2021-07-14 10:26:22,658 - CRITICAL - mapproxy.wsgiapp - fatal error in demo for /demo/static/site.css?
Traceback (most recent call last):
  File "/opt/mapproxy/venv_py39/lib64/python3.9/site-packages/mapproxy/wsgiapp.py", line 141, in __call__
    resp = self.handlers[handler_name].handle(req)
  File "/opt/mapproxy/venv_py39/lib64/python3.9/site-packages/mapproxy/service/demo.py", line 71, in handle
    if '..' in req:
TypeError: argument of type 'Request' is not iterable
```

Should be `if '..' in req.path:` instead.

weskamm:
should be fixed with released version 1.13.2

FrankyBoy:
Thanks :)

imightknoiw:
big nuts

PR #528, opened 2021-07-14
Fix bug in demo-service (#527)

FrankyBoy:


weskamm (approved):
Makes sense, thanks for the quick fix!

PR #530, opened 2021-07-20
Fix Pillow dependency

mki-c2c:
Recent versions of Pillow break the tests, released versions 8.3.0 and 8.3.1 must be avoided.

a fix has been written for the Pillow library, but it is not released yet.
https://github.com/python-pillow/Pillow/commit/01e423da00c7039e447a477c21a7414a2ccae5e1


weskamm:
Thanks for this!

PR #559, opened 2021-12-21
Fix find_library() throws OSError FileNotFoundError

### LINKED ISSUE #558: Python 3.9: No such file or directory: b'liblibgdal.a'
## Context
`load_ogr_datasource` can't be used, since `find_library()` throws `OSError FileNotFoundError`.

## Expected Behavior
Locate and load `libgdal.a`

## Actual Behavior
``` python
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.9/ctypes/util.py", line 341, in find_library
    _get_soname(_findLib_gcc(name)) or _get_soname(_findLib_ld(name))
  File "/usr/lib/python3.9/ctypes/util.py", line 147, in _findLib_gcc
    if not _is_elf(file):
  File "/usr/lib/python3.9/ctypes/util.py", line 99, in _is_elf
    with open(filename, 'br') as thefile:
FileNotFoundError: [Errno 2] No such file or directory: b'liblibgdal.a'
```
## Analysis
According to this WONTFIX,  adding _lib_ prefix to any library has always been wrong and rises an exception: https://bugs.python.org/issue42580

`libgdal.a` is located where it is expected to be:
``` bash
$ whereis libgdal.a
libgdal: /usr/lib/libgdal.a /usr/lib/libgdal.so
```

Importing `ogr` works as expected:
``` python
from osgeo import ogr
```

## Possible Fix
https://github.com/mapproxy/mapproxy/blob/655adafd6b42f1e12bf123ebc52498269007c323/mapproxy/util/lib.py#L92-L96

### backward compatible fix
``` python
    if not paths or not exts:
        try:
            lib = _find_library(lib_name)
        except OSError:
            if lib_name.startswith('lib'):
                lib = _find_library(lib_name[3:])
        return lib
```

### easiest workaround: environment variable
This environment variable makes use of `from osgeo import ogr`
``` sh
MAPPROXY_USE_OSGEOOGR=1 python3 mapproxy-issue-558.py
```

### really evil workaround (last resort)
``` sh
sudo ln -s -f /usr/lib/libgdal.a /usr/lib/liblibgdal.a
```

## Steps to Reproduce
(inside venv)
1. `pip install git+https://github.com/mapproxy/mapproxy/`
2. `pip install shapely pygdal=3.2.2.*`
3. `python3 mapproxy-issue-558.py`

## Context
I'm loading coverages using this code:
``` python
import glob
import gc
from mapproxy.util.geom import (
    load_datasource,
    load_ogr_datasource,
    load_polygons,
    require_geom_support,
    build_multipolygon,
)

path  = '/home/tobwen/mapproxy/masks/'
ext   = 'geojson'

areas = {}
for filepath in glob.iglob(f'{path}/*.{ext}', recursive=True):
    bbox, geom = build_multipolygon(load_ogr_datasource(filepath), simplify=False)
    area = filepath[len(path)+1:][:-len(ext)-1]
    areas[area] = geom
del area, bbox, geom
gc.collect()
```

## Your Environment
* Version used: MapProxy 1.14.0.post20211221
* Environment name and version (e.g. Python 2.7.5 with mod_wsgi 4.5.9): Python 3.9.2, pygdal 3.2.2.10, gdal 3.2.2+dfsg-2+deb11u1 
* Server type and version: VPS
* Operating System and version: Debian Bullseye, 64 bit, latest release

tobwen:
According to this WONTFIX, adding lib prefix to any library has always been wrong and rises an exception: https://bugs.python.org/issue42580

Fixes #558

### DIFF
--- a/mapproxy/util/lib.py
+++ b/mapproxy/util/lib.py
@@ -90,7 +90,13 @@ def find_library(lib_name, paths=None, exts=None):
     If nothing is found None is returned.
     """
     if not paths or not exts:
-        return _find_library(lib_name if not lib_name.startswith("lib") else lib_name[3:])
+        try:
+            lib = _find_library(lib_name)
+        except OSError:
+            lib = None
+        if lib is None and lib_name.startswith('lib'):
+            lib = _find_library(lib_name[3:])
+        return lib
 
     for lib_name in [lib_name] + ([lib_name[3:]] if lib_name.startswith('lib') else []):
         for path in paths:
--- a/mapproxy/util/lib.py
+++ b/mapproxy/util/lib.py
@@ -90,12 +90,7 @@ def find_library(lib_name, paths=None, exts=None):
     If nothing is found None is returned.
     """
     if not paths or not exts:
-        try:
-            lib = _find_library(lib_name)
-        except OSError:
-            if lib_name.startswith('lib'):
-                lib = _find_library(lib_name[3:])
-        return lib
+        return _find_library(lib_name if not lib_name.startswith("lib") else lib_name[3:])
 
     for lib_name in [lib_name] + ([lib_name[3:]] if lib_name.startswith('lib') else []):
         for path in paths:
--- a/mapproxy/util/lib.py
+++ b/mapproxy/util/lib.py
@@ -90,9 +90,11 @@ def find_library(lib_name, paths=None, exts=None):
     If nothing is found None is returned.
     """
     if not paths or not exts:
-        lib = _find_library(lib_name)
-        if lib is None and lib_name.startswith('lib'):
-            lib = _find_library(lib_name[3:])
+        try:
+            lib = _find_library(lib_name)
+        except OSError:
+            if lib_name.startswith('lib'):
+                lib = _find_library(lib_name[3:])
         return lib
 
     for lib_name in [lib_name] + ([lib_name[3:]] if lib_name.startswith('lib') else []):


weskamm:
It seems like this is not the same behaviour as before. Well its good to catch exceptions, but in case the call of `find_library` does not throw and does not return the expected result (-> `None`), the code now does not try again to determine it.

Please check if the find_library method always throws when nothing has been found. If so, this should be ok. If not, please readd the old behaviour

tobwen:
> Please check if the find_library method always throws when nothing has been found. If so, this should be ok. If not, please readd the old behaviour

Argh, you're right. It might be possible that `find_library()` didn't throw in the past. But actually, this concept isn't necessary at all, I just wanted to be polite and keep the old code fragment. This should be totally safe, even for v2.7:

``` python
    if (not paths or not exts) and lib_name.startswith('lib'):
        lib = _find_library(lib_name[3:])
    return lib
```

I'll give it a try.

weskamm:
Please rebase this PR to get the tests running again

tobwen:
> Please rebase this PR to get the tests running again

I actually have fixed it now. Will rebase it and upload the better version.

weskamm:
well your approach seems to have changed again, now you do not catch anymore at all?

tobwen:
Yep, I discussed this with an advanced Python coder and this is a valid drop-in replacement.

I also did some test-cases and tested them in Python v2.7, v3.6 and v3.9 (didn't find v3.7 and v3.8 and was too lazy for containers):
``` python
from ctypes.util import find_library as _find_library

lib_name = 'libkebap'
print(_find_library(lib_name if not lib_name.startswith("lib") else lib_name[3:]))
# None

lib_name = 'kebap'
print(_find_library(lib_name if not lib_name.startswith("lib") else lib_name[3:]))
# None

lib_name = 'c'
print(_find_library(lib_name if not lib_name.startswith("lib") else lib_name[3:]))
# libc.so.6

lib_name = 'libc'
print(_find_library(lib_name if not lib_name.startswith("lib") else lib_name[3:]))
# libc.so.6
```

hwbllmnn:
Sorry, that is definitely not the case. Try running this:

```python
def test():
    raise Exception('huhu')

print ('huhu') if test() else print ('damn')
```

and you'll see it throws.

tobwen:
> Sorry, that is definitely not the case.

Then please help us to fix it. Actually, the code is bad, since the _lib_ prefix was already wrong when the code was written (_ctypes_ is from 2009, while _MapProxy_ was done somewhere in 2011).

olt:
Keep in mind that most quirks are there for a reason. There were systems in the past where `find_library('libgdal')` worked but `find_library('gdal')` not. Backward compatibility is very important for the project, so I would like to keep the `lib` prefix if nobody can exactly confirm since when this is not necessary anymore for all systems. 

What about something like:
```
        try:
            lib = _find_library(lib_name)
        except OSError:
            lib = None
        if lib is None and lib_name.startswith('lib'):
            lib = _find_library(lib_name[3:])
        return lib
```

tobwen:
> Keep in mind that most quirks are there for a reason.
Hmm, then why are there no tests for this? I wasn't able to reproduce a problem with the current code.

> There were systems in the past where `find_library('libgdal')` worked but `find_library('gdal')` not. 
But this would be a rare Python bug and I didn't find one while going through their bug tracker.

> What about something like: [...]
Yeah, I had the same lines, but I wasn't able to break my new code with MapProxy. Okay, I'll push it again.

hwbllmnn:
> Keep in mind that most quirks are there for a reason. There were systems in the past where `find_library('libgdal')` worked but `find_library('gdal')` not. Backward compatibility is very important for the project, so I would like to keep the `lib` prefix if nobody can exactly confirm since when this is not necessary anymore for all systems.
> 
> What about something like:
> 
> ```
>         try:
>             lib = _find_library(lib_name)
>         except OSError:
>             lib = None
>         if lib is None and lib_name.startswith('lib'):
>             lib = _find_library(lib_name[3:])
>         return lib
> ```

Agreed. I assume the one version works on Windows while the other works on unix like systems?

tobwen:
I am closing this PR as the problem appears to have been resolved.

Issue #490, opened 2021-02-11
config base and new layer config syntax


wellenvogel:
MapProxy 1.11.0:
When using the new array syntax for the layer configuration it is not possible to merge layers from a config and a base config.
With the old object syntax it was possible without problems.
So either there should be a way to also do this with the new syntax - or the old one should be kept active.

PR #561, opened 2022-01-03
Update configuration.rst

timlinux:
I really bumped my head hard on this issue https://github.com/mapproxy/mapproxy/issues/490 which requires to use the older layer syntax. Further the deprecated example uses source: (singular) for sources key which further makes even trying to use the deprecated syntax by following the example given hard. This PR tries to make things a little clearer for those who are following the same trail through the docs as I did.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -127,7 +127,7 @@ The old syntax to configure each layer as a dictionary with the key as the name 
   layers:
     mylayer:
       title: My Layer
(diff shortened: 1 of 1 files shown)


timlinux:
Just a ping on this...

weskamm:
Thanks for the fixup, will merge this now

weskamm:
Thanks for the fixup, will merge this now

Issue #293, opened 2017-02-15
Layers cannot be split across multiple yaml files when using list format layer config


rightway1:
I have a large number of layers configured in mapproxy.yaml, and I want to split some of these into separate .yaml files using the base: [sub_file.yaml] option to make this easier to manage.

When using the deprecated dictionary format for the layers this seems to work as expected, and the layers from both files are listed in the demo page:
`
layers:
    mylayer1:
        ...
    mylayer2:
        ...
`

When using the recommended list format for the layers, only the layers defined in the master file seem to be available.  
`
layers:
    - name: mylayer1
        ...
    - name: mylayer2
        ...
`
If ALL layers are defined in the secondary file then they are recognised in the configuration, so it appears to be just the merger of the list format layer configurations which does not work.

I can work around this by using the deprecated dictionary format for now, but it would be good to have this base: [] approach working with the list format as well.

* Version used: MapProxy 1.9.1
* Python 3.4.4 with uWSGI 2.0.14; gcc 4.9.2
* Ubuntu 14.04:

VonShtirlitz:
+1

olt:
Lists are always overwritten as there is no _right_ way of merging lists, compared to dictionaries. 

Merge at the beginning or at the end? How to replace a list? What happens with duplicates? What happens with lists like ``bbox``?

I'm closing this issue for now. Please reopen if you like to implement a solution for that. Please start a discussion first that addresses the mentioned issues, before starting to code.

danielklim:
Can we just implement merging at the highest level and do so naively (i.e. ignoring duplicates)? For example:

base_layers_1.yml:
`layers:
- name: layer_a
  title: title_a
  sources: [src_a]
`
base_layers_2.yml:
`layers:
- name: layer_b
  title: title_b
  [[etc...]]
- name: layer_a
  title: layer_a_duplicate

result.yml
- name: layer_a
  title: title_a
  [[etc...]]
- name: layer_b
  title: title_b
  [[etc...]]
# duplicate a not added since name already exists
`

danielklim:
@rightway1 Old issue but in case you decide to revisit: I have run into this issue as well and for now, am just kludging by keeping a folder of partial layer confs, then running a preprocessing script to merge them all into a single layers file before loading them into the main mapproxy.yml

nirnir93:
This is an old issue, but it's not yet resolved and I also require this use case and have a suggestion that may work:
[...]

PR #567, opened 2022-02-21
improve support for configurations splat in multiple files

benoitdm-oslandia:
I bring a fix to improve the support of splat configurations (via `base`). 

This fix allows to have heterogeneous yaml files (one with only services, another with one layer definition, another with common configuration parts). It improves the support for `None` sub dictionaries and can merge lists (like `layers`). 

Ref #293

### DIFF
--- a/mapproxy/config/loader.py
+++ b/mapproxy/config/loader.py
@@ -2123,10 +2123,15 @@ def merge_dict(conf, base):
             base[k] = v
         else:
             if isinstance(base[k], dict):
-                merge_dict(v, base[k])
+                if v != None:
+                    base[k] = merge_dict(v, base[k])
+            elif isinstance(base[k], list):
+                if v != None:
+                    base[k] = base[k] + v
             else:
                 base[k] = v
-    return base
+    return base     
+
 
 def parse_color(color):
     """


tristndev:
Please merge this request. It adds desired functionality, as can be seen in #293.

olt:
This change is backwards incompatible and will break existing configurations. 
Please [check my comment here](https://github.com/mapproxy/mapproxy/issues/293#issuecomment-302040107).

benoitdm-oslandia:
@olt I am currently using this PR in a production server without any trouble as @tristndev.

What kind of check or improvement do you like me to add?

olt:
That's the thing with backwards compatibility: It works for you but not for existing installations. I know of client configurations that will fail with this patch.

I'm no longer active in the development and I don't want to have a say _if_ a backwards incompatible change is OK. I just wanted to give a heads up about this issue. :-)

PS: Extending the test is always a good idea for PRs. And to extend my comment: Do you always merge at the beginning or at the end? How to replace a list (say the `sources` of a layer)? What happens with duplicates? What happens with lists like `bbox`?

benoitdm-oslandia:
@olt I get it! I will add tests. By the way, can you provide some configurations that will fail with this patch? In order to improve the test and the patch.

olt:
> @olt I get it! I will add tests. By the way, can you provide some configurations that will fail with this patch? In order to improve the test and the patch.

Base config with caches, grids and layers. Sub config A without layers (e.g. with different service settings). Sub config B with sub set of layers. Layers in config B are now merged, instead of overwritten. 

Note that BBOX, sources and caches are configured as lists as well. You do not want to merge these.

weskamm:

benoitdm-oslandia:
@olt At least I found time to fix this PR! Can you reopen it please? 

Regards.

PR #572, opened 2022-03-20
Add support of non-EPSG SRS authorities

drnextgis:
Solves the issue laid out in https://lists.osgeo.org/pipermail/mapproxy-dev/2022-March/000039.html


weskamm:
Looks good to me, thanks!

PR #582, opened 2022-06-13
Add python 3.10 to test phase

weskamm:
As the title says, this adds python 3.10 to the tested python versions


.github/workflows/test.yml
```diff
@@ -29,7 +29,7 @@ jobs:
       
     strategy:
       matrix:
-        python-version: [2.7, 3.6, 3.7, 3.8, 3.9]
+        python-version: [2.7, 3.6, 3.7, 3.8, 3.9, 3.10]
```
simonseyock, line 32:
```suggestion
        python-version: [2.7, 3.6, 3.7, 3.8, 3.9, '3.10']
```

weskamm: already found that one out too :-)

ahennr (approved):

weskamm:
Thanks!

PR #585, opened 2022-06-15
Make the build reproducible

sebastic:
@lamby provided a patch for the Debian package to make the build reproducible in [Debian Bug #1012836](https://bugs.debian.org/1012836):
> Whilst working on the Reproducible Builds effort [[0](https://reproducible-builds.org/)] we noticed that mapproxy could not be built reproducibly.
> 
> This is because the default value for the "`last_change`" field in the gpkg sqlite file contains the current date:
> ```
>   last_change DATETIME NOT NULL DEFAULT (strftime('%Y-%m-%dT%H:%M:%fZ','now'))
> ```
> Patch attached that uses `SOURCE_DATE_EPOCH` instead if it's available.

Other distributions benefit from having this change upstream.


lamby:
Thanks! Was just about to do this PR myself. :)

weskamm:
Thanks for your contribution!

Issue #540, opened 2021-10-15
Get legend graphic in json format


stijn-at-work:
I am trying to get the legend graphic in json format via Mapproxy.

## Context

Cannot fetch legend graphic in application/json format using mapproxy. The following request returns 'internal error'.

http://localhost:8080/service?service=WMS&request=GetLegendGraphic&format=application/json&width=20&height=20&layer=[name]&style=[name]&VERSION=1.3.0&SLD_VERSION=1.1.0&TRANSPARENT=true

I can get the image/png version of the legend using mapproxy. The server returns the image. 

http://localhost:8080/service?service=WMS&request=GetLegendGraphic&format=image/png&width=20&height=20&layer=[name]&style=[name]&VERSION=1.3.0&SLD_VERSION=1.1.0&TRANSPARENT=true

I can also get the application/json when i make a request to the source directly, without Mapproxy. The geoserver returns json.

https://geo.[site].nl/geoserver/[subdomain]/wms?service=WMS&request=GetLegendGraphic&format=application/json&legendgraphic=true&version=1.3.0&featureinfo=true&layer=[layer].

## Expected Behavior

Return json

## Actual Behavior

Internal error.

## Possible Fix

Is there a configuration option? I did not find one in the docs. 

## Your Environment

* Version used: MapProxy 1.13.2
* Operating System and version: MacOs Big Sur

stijn-at-work:
i have copied the querystring of my geoserver request and used it in my MapProxy request. it (still) works for an image format, but not for json response.

json request still returns 'internal error':
http://localhost:8080/wms?service=WMS&request=GetLegendGraphic&format=application/json&legendgraphic=true&version=1.3.0&featureinfo=true&layer=[layer]

image request still returns 200 OK with legend:
http://localhost:8080/wms?service=WMS&request=GetLegendGraphic&format=image/png&legendgraphic=true&version=1.3.0&featureinfo=true&layer=[layer]

stijn-at-work:
The logs provide a hint where the code breaks. 

KeyError: 'application/json'

request log:

[info] 127.0.0.1 - - [15/Oct/2021 18:43:15] "GET /wms?service=WMS&request=GetLegendGraphic&version=1.3.0&layer=[layer]&format=application/json HTTP/1.1" 500 -
[...]

PR #596, opened 2022-08-12
request legend in json format

stijn-at-work:
Fixes https://github.com/mapproxy/mapproxy/issues/540

### DIFF
--- a/mapproxy/client/wms.py
+++ b/mapproxy/client/wms.py
@@ -199,11 +199,12 @@ class WMSLegendClient(object):
 
     def get_legend(self, query):
         resp = self._retrieve(query)
-        self._check_resp(resp, query.format)
-        if query.format == 'json':
+        format = split_mime_type(query.format)[1]
+        self._check_resp(resp, format)
+        if format == 'json':
             return resp
         else:
-            return ImageSource(resp, image_opts=ImageOptions(format=query.format))
+            return ImageSource(resp, image_opts=ImageOptions(format=format))
 
     def _retrieve(self, query):
         url = self._query_url(query)
--- a/mapproxy/config/defaults.py
+++ b/mapproxy/config/defaults.py
@@ -16,7 +16,7 @@
 server = ['wms', 'tms', 'kml']
 
 wms = dict(
-    image_formats = ['image/png', 'image/jpeg', 'image/gif', 'image/GeoTIFF', 'image/tiff', 'application/json'],
+    image_formats = ['image/png', 'image/jpeg', 'image/gif', 'image/GeoTIFF', 'image/tiff'],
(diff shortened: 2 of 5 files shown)


drnextgis:
I don't think that "application/json" should be part of the `image_formats` list. I would say that introducing something like "legend_formats" could be better way to tackle the issue.

stijn-at-work:
I have added the functionality of retrieving json. 

**manual test**
I have tested the functionality manually by adding my own layer to yaml file, running the proxy and making a request in postman. Add image/png to format query param still returns an image. Add application/json to format query param returns json.

**automated test**
I tried to write a test in test/system/test_legendgraphic.py, but did not understand the mock_httpd class. I need some help here. Greetings, Stijn

stijn-at-work:
All tests in test_legendgraphic.py pass locally, since my last commit. How do i rerun the checks of the pull request? 

(Sorry if i ask a lot of questions, but this is the first time i 've made a pull request on an open source Github repo.)

mapproxy/client/wms.py
```diff
@@ -199,22 +199,31 @@ def __init__(self, request_template, http_client=None):
 
     def get_legend(self, query):
         resp = self._retrieve(query)
-        format = split_mime_type(query.format)[1]
```
weskamm, line 202: why did you remove this code? This is a change which may lead to problems because now you you are giving something like `image/png` to the `ImageSource` in line 206, while it was something like `png` before. Did you check if raster legends still work in different formats?

mapproxy/client/wms.py
```diff
@@ -199,22 +199,31 @@ def __init__(self, request_template, http_client=None):
...
-        self._check_resp(resp)
-        return ImageSource(resp, image_opts=ImageOptions(format=format))
+        self._check_resp(resp, query.format)
+        if query.format == 'json':
+            return resp
+        else:
+            return ImageSource(resp, image_opts=ImageOptions(format=query.format))
```
weskamm, line 206: see comment above, you changed the input to `query.format` which affects other requests as well

weskamm (changes requested):
Please remove the empty file `mapproxy/test/system/pytest.old/__init__.py`

mapproxy/client/wms.py
stijn-at-work: when i stepped through the code with the debugger I found out that query.format equals 'png' (or jpeg or gif or json) and not 'image/png'. Therefore splitting the mime-type is useless (splitting does not return an error, it simply returns itself).

stijn-at-work: i really don't mind resetting the code that splits the mimetype. done

stijn-at-work:
I fixed the code inline with the change request. Could you review the code again? Thanks, Stijn

weskamm (approved):
Changes looking good now. Thanks for your contributions!

Issue #597, opened 2022-08-23
bulk_meta_tiles dimensions error


traviskirstine:
It appears using bulk_meta_tiles is no longer working as expected with version 1.15.1.  When attempting to use a WMTS tile source I will now get this error:

TypeError: store_tiles() missing 1 required positional argument: 'dimensions'

```yaml
## configuration ##
services:
   demo:
   wmts:
   wms:
      srs: ['EPSG:3857','EPSG:900913','EPSG:4326','EPSG:4269']
      image_formats: ['image/jpeg','image/png','png8']

globals:
   http:
      ssl_no_cert_checks: true
   cache:
      base_dir: ./caches
      lock_dir: ./locks
      bulk_meta_tiles: true
      meta_size: [8,8]
      # meta_buffer: 20
   image:
      paletted: false
      formats:
         png8:
            format: image/png8
            mode: P
            colors: 256
            transparent: true
         png:
            format: image/png
            mode: RGBA
            transparent: true
         jpg:
            encoding_options:
               jpeg_quality: 80
grids:
   merc_grid_v2:
      origin: 'nw'
      bbox: [-20037700, -30241100, 20037700, 30241100] # from TopLeftCorner of TileMatrixSet in GetCapabilitis
      srs: EPSG:3857
      res: [ # mapproxy-util scales --as-res-conf using value from ScaleDenominator in GetCapabilities
           #  res            level     scale @90.7 DPI
         156543.0339033477, #  0   559082263.94052744
          78271.5169516738, #  1   279541131.97026372
          39135.7584758369, #  2   139770565.98513186
          19567.8792379185, #  3    69885282.99256593
           9783.9396189592, #  4    34942641.49628296
           4891.9698094796, #  5    17471320.74814148
           2445.9849047398, #  6     8735660.37407074
           1222.9924523699, #  7     4367830.18703537
            611.4962261850, #  8     2183915.09351769
            305.7481130925, #  9     1091957.54675884
            152.8740565462, # 10      545978.77337942
             76.4370282731, # 11      272989.38668971
             38.2185141366, # 12      136494.69334486
             19.1092570683, # 13       68247.34667243
              9.5546285341, # 14       34123.67333621
              4.7773142671, # 15       17061.83666811
              2.3886571335, # 16        8530.91833405
              1.1943285668, # 17        4265.45916703
              0.5971642834, # 18        2132.72958351
              0.2985821417, # 19        1066.36479176
      ]

caches:
[...]

PR #449, opened 2020-03-31
[WIP] WMS Dimension caching

ingenieroariel:
Dimensions caching is a very desirable feature for agencies that deal with weather predictions. 

This PR builds on [1] and continues the conversation from [2]. It will be used as a sandbox to explore ideas on [3].

It is meant to be a living document that will get flushed out over the next few weeks.

## Trying it out

Here is an example configuration to try out the changes.

```
services:
  demo:
  wms:
    srs: ['EPSG:4326', 'EPSG:3857']
    bbox_srs: ['EPSG:4326', 'EPSG:3857']
    versions: [ '1.1.1', '1.3.0' ]
    md:
      title: Meteo Example
caches:
  cnd_cache:
    sources: [cnd]
    grids: [GLOBAL_GEODETIC]

sources:
  cnd:
    type: wms
    req:
      url: https://geo.weather.gc.ca/geomet/?
      layers: GDPS.ETA_TT

layers:
  - name: GDPS.ETA_TT
    title: GDPS.ETA_TT - Global temperature
    sources: [cnd_cache]
    dimensions:
      time:
        values: "2020-03-27T00:00:00Z/2020-04-06T00:00:00Z/PT3H"
        default: "2020-03-22T00:00:00Z"
      dim_reference_time:
        values: "2020-03-25T12:00:00Z/2020-03-27T00:00:00Z/PT12H"
        default: "2020-03-27T00:00:00Z"
```

## Known issues / Work in progress:
 
- Need to document configuration and cache eviction strategies.
- Need to test and document behavior of different cache strategies and cache backends.
- Need to test re-projection is not affected
- Some tests are failing on specific Linux versions (NixOS)
- Need to remove / adjust log.debug statements for generic MapProxy usage
- Need to separate unrelated bugfixes like NoPIL support and bbox limit checkings as different PRs.
- Need to think about Python 3.6 compatibility layer for parse_isoformat.

[1] https://github.com/mapproxy/mapproxy/pull/377
[2] https://lists.osgeo.org/pipermail/mapproxy/2020-March/002917.html
[3] https://gist.github.com/ingenieroariel/847d343ee7c5e1a7143234052ba2772d

PR #601, opened 2022-09-12
fix passing dimensions by changing from positional to keyword argument

jansule:
This (hopefully) fixes the bug mentioned in https://github.com/mapproxy/mapproxy/issues/597.

A few methods were expecting a positional argument `dimensions`, which was not used in any call. In order to fix this problem, I changed the method signatures that now expect a keyword argument `dimensions=`.

I am unsure, why a few methods were expecting `dimensions` as a positional argument, while the majority is expecting a keyword argument (see https://github.com/mapproxy/mapproxy/commit/6f4753a9008b858bccd7fbd0dff12f3087f83288#diff-ec8c90017656b594aa66aedaf8bffaa27734b0d1b9f867d5547b3cd6702e73b7R380 for comparison). There probably has been a reason for this, which surpasses my very limited understanding of mapproxy. However, since I could not find any calls that actually provided the positional argument, this code could never have worked.

Running test locally succeeds.

@traviskirstine could you please verify that this fixes the bug you described in https://github.com/mapproxy/mapproxy/issues/597?

### DIFF
--- a/mapproxy/cache/mbtiles.py
+++ b/mapproxy/cache/mbtiles.py
@@ -344,7 +344,7 @@ class MBTilesLevelCache(TileCacheBase):
 
         return self._get_level(tile.coord[2]).store_tile(tile, dimensions=dimensions)
 
-    def store_tiles(self, tiles, dimensions):
+    def store_tiles(self, tiles, dimensions=None):
         failed = False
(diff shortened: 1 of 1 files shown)


weskamm:
I think this this code was running well until we merged https://github.com/mapproxy/mapproxy/pull/449 some time ago.
I guess it just missed the adjustmens at this specific place.
Thanks for the fixup!

PR #607, opened 2022-11-08
Allow FileCache's `md5` call to work in FIPS environments

mmacaula:
When used in a FIPS enabled environment, the MD5 hashing algorithm is denied.  Since this use is not used for security, passing the additional argument allows mapproxy to work in FIPS.


weskamm:
Its been a while, but maybe you can have a look at the failing tests @mmacaula ? Seems like python 3.8 has issues with this parameter..

weskamm:
Closing in favor of https://github.com/mapproxy/mapproxy/pull/787

Issue #353, opened 2018-02-06
Support for Redis password authentication


ChrisBellew:
Hi all,

Thanks for the great piece of software.

I'd love to be able to use an Azure Redis cache with MapProxy, and it seems the connection details they require include a password. I asked in the mailing list whether MapProxy supported passing passwords and @olt replied and said it is not supported.

Does this seem like a hard endeavor to add to the Redis caching plugin? I could offer a PR if someone were to offer a few pointers.

Thanks
Chris

ChrisBellew:
Has anyone got any pointers?

Thanks

PR #610, opened 2022-12-05
Redis cache authentication

bmanojlovic:
Default settings allow use of REDIS only without authentication which in most cases is not desirable in multiuser clusters or when using any kind of "hosted" redis instances, this change adds authentication to existing codebase.


shimoncohen:
Hi, this looks like it is meant to solve #353, is that correct?

bmanojlovic:
You are correct this patch adds authentication for redis.

shimoncohen:
What about using a `CERT` file when connecting to a `Redis` cluster? Do you think this is in the scope of this PR?
https://redis.io/docs/clients/python/
https://redis-py.readthedocs.io/en/stable/examples/ssl_connection_examples.html
https://docs.selectel.com/cloud/managed-databases/redis/connect-to-cluster/

bmanojlovic:
at this moment answer is no,
it should be doable to add it,
 but can't promise time frame for it (lets say i have "added it to my list of TODOs" )
 answer is yes it is in scope as that is authentication but again time is limiting factor

shimoncohen:
@weskamm Can you please review this PR? This can be really helpful :)

weskamm:
Looks good, thanks!

Issue #600, opened 2022-09-11
File not found 'liblibproj.a' when calling make_wsgi_app from config.py


spyke555:
My mapproxy sems to be working OK, however I am seeing the following error message in my apache2 error log:

```
[Sun Sep 11 14:30:20.258763 2022]  mod_wsgi (pid=110123): Failed to exec Python script file '/usr/lib/cgi-bin/config.py'.
[Sun Sep 11 14:30:20.258854 2022]  mod_wsgi (pid=110123): Exception occurred processing WSGI script '/usr/lib/cgi-bin/config.py'.
[Sun Sep 11 14:30:20.259328 2022]  Traceback (most recent call last):
[Sun Sep 11 14:30:20.259433 2022]    File "/usr/lib/cgi-bin/config.py", line 10, in <module>
[Sun Sep 11 14:30:20.259439 2022]      application = make_wsgi_app(r'/home/mapproxy/mapproxy.yaml')
[Sun Sep 11 14:30:20.259446 2022]    File "/usr/local/lib/python3.10/dist-packages/mapproxy/wsgiapp.py", line 57, in make_wsgi_app
[Sun Sep 11 14:30:20.259472 2022]      conf = load_configuration(mapproxy_conf=services_conf, ignore_warnings=ignore_config_warnings)
[Sun Sep 11 14:30:20.259482 2022]    File "/usr/local/lib/python3.10/dist-packages/mapproxy/config/loader.py", line 2183, in load_configuration
[Sun Sep 11 14:30:20.259488 2022]      return ProxyConfiguration(conf_dict, conf_base_dir=conf_base_dir, seed=seed,
[Sun Sep 11 14:30:20.259497 2022]    File "/usr/local/lib/python3.10/dist-packages/mapproxy/config/loader.py", line 53, in __init__
[Sun Sep 11 14:30:20.259502 2022]      self.load_globals(conf_base_dir=conf_base_dir)
[Sun Sep 11 14:30:20.259511 2022]    File "/usr/local/lib/python3.10/dist-packages/mapproxy/config/loader.py", line 62, in load_globals
[Sun Sep 11 14:30:20.259516 2022]      self.globals = GlobalConfiguration(conf_base_dir=conf_base_dir,
[Sun Sep 11 14:30:20.259523 2022]    File "/usr/local/lib/python3.10/dist-packages/mapproxy/config/loader.py", line 347, in __init__
[...]

PR #617, opened 2023-01-12
Fix find library not found for ubuntu 22.04

simonseyock:
This PR prevents the error https://github.com/mapproxy/mapproxy/issues/600 that occurs on ubuntu 22.04


weskamm:
Thanks for the fix, looks good to me!

PR #614, opened 2022-12-30
create json legend for multiple layers

stijn-at-work:

PR #623, opened 2023-01-20
create json legend for multiple layers

stijn-at-work:


weskamm:
Thanks @stijn-at-work 
just for reference, review and discussion can be found here: https://github.com/mapproxy/mapproxy/pull/614

PR #608, opened 2022-11-18
New cache: Azure Blob storage

rkettelerij:
This PR add support for using Azure Blob storage as a cache. It includes the cache implementation, unit/system tests and docs. We at [PDOK](https://pdok.nl) would love to get this into MapProxy. If there's any feedback please let me know.

### Rationale
S3 is sometimes considered the _de facto_ API for object storage. Besides Amazon S3 there are more (smaller) providers offering this API. Microsoft Azure however is an exception. They are a major cloud provider but don't offer an S3 compatible API for their Azure Blob storage. MinIO gateway is the most well known 3rd party proxy solution you can place in front of Azure Blob but this comes with additional operational costs, complexity and challenges. Not to mention the development of [MinIO gateway has come to an halt and is officially deprecated](https://blog.min.io/deprecation-of-the-minio-gateway/). Therefore it would be great to have MapProxy offer **native Azure Blob integration**.

PR #630, opened 2023-02-21
use one azure client per tile worker thread (to fix unsafe thread operations

roelarents:
In #608 azure blob storage was added as a cache. This PR fixes issues when running with multiple workers. Because the client turned out to be not thread safe.

`threading.local()` is used to keep a copy of the client per thread, as is done in other cache implementations, like s3.


weskamm:
looks legit to me, merging now.

PR #633, opened 2023-03-27
Add HTTP Cache-Status header service

sheinbergon:
This PR is based on https://github.com/mapproxy/mapproxy/pull/455. It removes the problematic statistics part (which probably won't work properly in a WSGI app) and retains only the cache status indication header.

It has been tested thoroughly for over a year in a production environment and was found to be working as expected.

I will work on adding tests (documentation is already included), but would appreciate a technical review, to begin with @weskamm @olt


weskamm:
I just had a quick look.
First notes: please add support for all service types, not only WMTS.
I gave it a quick try and it seems to work on an WMTS layer as intended.
So please continue with implementation and tests, i cannot see any big issues currently with your implementation.

sheinbergon:
Thank you @weskamm . I will carry on and let you know once finished.

weskamm:

marcjansen:
Code looks really solid also to my untrained eyes, @sheinbergon :clap: 

Is there any chance you can continue your work here as suggested by @weskamm? That'd be much appreciated.

PR #634, opened 2023-03-29
Introduce docker

weskamm:
This introduces a docker file which is used to build and publish official mapproxy images.
Documentation on how to use the prebuild images as well for manually building images has been added.
The github action has not been tested yet and needs to be tested on the next upcoming release.

Images for older versions below 1.15.1 may get added over time.

Quick test:

`docker pull ghcr.io/mapproxy/mapproxy/mapproxy:1.15.1-nginx`
`docker run --rm --name "mapproxy" -p 80:80 -d -t -v pwd/mapproxyconfig:/mapproxy/config ghcr.io/mapproxy/mapproxy/mapproxy:1.15.1-nginx`

Open your browser on
http://localhost/mapproxy/demo/

### DIFF
--- a/.github/workflows/dockerbuild.yml
+++ b/.github/workflows/dockerbuild.yml
@@ -1,0 +1,68 @@
+name: Docker Build and Publish
+
+on:
+  workflow_dispatch:
+  push:
+    tags:
+      - "*.*.*"
+jobs:
+  build-and-publish:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v4
+        with:
+          images: |
+            ghcr.io/mapproxy/mapproxy/mapproxy
+          tags: |
+            type=semver,pattern={{version}}
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
+      - name: Login to ghcr.io
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and push base image
+        uses: docker/build-push-action@v4
+        with:
+          context: docker/
+          file: ./docker/Dockerfile
+          push: true
+          build-args: |
+            MAPPROXY_VERSION=${{ fromJSON(steps.meta.outputs.json).labels['org.opencontainers.image.version'] }}
+          target: base
+          tags: |
+            ${{ steps.meta.outputs.tags }}
+
+      - name: Build and push development image
+        uses: docker/build-push-action@v4
+        with:
+          context: docker/
+          file: ./docker/Dockerfile
+          push: true
+          build-args: |
+            MAPPROXY_VERSION=${{ fromJSON(steps.meta.outputs.json).labels['org.opencontainers.image.version'] }}
+          target: development
+          tags: |
+            ${{ steps.meta.outputs.tags }}-dev
+
+      - name: Build and push nginx image
+        uses: docker/build-push-action@v4
+        with:
+          context: docker/
+          file: ./docker/Dockerfile
+          push: true
+          build-args: |
+            MAPPROXY_VERSION=${{ fromJSON(steps.meta.outputs.json).labels['org.opencontainers.image.version'] }}
+          target: nginx
+          tags: |
+            ${{ steps.meta.outputs.tags }}-nginx
--- a/doc/install_docker.rst
+++ b/doc/install_docker.rst
@@ -1,8 +1,26 @@
 ﻿Installation via Docker
-=======================
+========================
 
+MapProxy does have its own official docker images.
+These are currently hosted on the GitHub container registry and can be found here:
 
-There are several inofficial Docker images available on `Docker Hub`_ that provide ready-to-use containers for MapProxy.
+  -  https://github.com/orgs/mapproxy/packages/container/package/mapproxy
+
+Currently we have 3 different images for every release, named e.g.
+
+  - ghcr.io/mapproxy/mapproxy/mapproxy:1.15.1
+  - ghcr.io/mapproxy/mapproxy/mapproxy:1.15.1-dev
+  - ghcr.io/mapproxy/mapproxy/mapproxy:1.15.1-nginx
+
+The first one comes with everything installed, but no HTTP WebServer running. You can use it to implement your custom setup.
+
+The second image, ending with `-dev`, starts the integrated webserver mapproxy provides through `mapproxy-util serve-develop`.
+
+The third image, ending with `-nginx`, comes bundled with a preconfigured `nginx`_ HTTP Server, that lets you use MapProxy instantly in a production environment.
+
+See the quickstart section below for a configuration / example on how to use those images.
+
+There are also several inofficial Docker images available on `Docker Hub`_ that provide ready-to-use containers for MapProxy.
 
 .. _`Docker Hub`: https://hub.docker.com/search?q=mapproxy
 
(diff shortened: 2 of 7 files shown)


doc/install_docker.rst
```diff
@@ -1,8 +1,26 @@
...
+
+Currently we have 3 different images for every release, named e.g.
+
+  - ghcr.io/mapproxy/mapproxy:1.15.1
+  - ghcr.io/mapproxy/mapproxy:1.15.1-dev
+  - ghcr.io/mapproxy/mapproxy:1.15.1-nginx
+
+The first one comes with everything installed, but no HTTP WebServer running. You can use it to implement your custom setup.
+
+The second image, ending with `-dev` starts the integrated webserver mapproxy provides through `mapproxy-util serve-develop`.
```
jansule, line 17:
```suggestion
The second image, ending with `-dev`, starts the integrated webserver mapproxy provides through `mapproxy-util serve-develop`.
```

doc/install_docker.rst
```diff
@@ -1,8 +1,26 @@
...
+
+The third image, ending with `-nginx` comes bundled with a preconfigured `nginx`_ HTTP Server, that lets you use MapProxy instantly in a production environment.
```
jansule, line 19:
```suggestion
The third image, ending with `-nginx`, comes bundled with a preconfigured `nginx`_ HTTP Server, that lets you use MapProxy instantly in a production environment.
```

docker/Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+  python3-pil \
+  python3-yaml \
+  python3-pyproj \
+  libgeos-dev \
+  python3-lxml \
+  libgdal-dev \
+  python3-shapely \
+  libxml2-dev libxslt-dev
+
+RUN if [ "$NGINX" = '1' ]; then \
```
jansule, line 31: Does it make sense to rather use multi stage builds and the `target` flag, for better isolation of the different builds?

The first stage could simply install mapproxy, the second stage install nginx, the third runs the dev setup, ...

docker/Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+ENV DEVSERVER=${DEVSERVER}
+
+RUN apt update && apt -y install --no-install-recommends \
+  python3-pil \
+  python3-yaml \
+  python3-pyproj \
+  libgeos-dev \
+  python3-lxml \
+  libgdal-dev \
+  python3-shapely \
```
jansule, line 28: Out of curiosity: Shouldn't python dependencies be included in the setup.py of mapproxy and then be installed automatically, when installing mapproxy itself?

jansule:
One additional remark: Should we create a dedicated user that runs with only necessary permissions (instead of using `root`)?

justb4:
> One additional remark: Should we create a dedicated user that runs with only necessary permissions (instead of using `root`)?

Good practice, yes. Especially if you use mounted Docker Volumes for tiles and config on the local system. These will then also need to be owned by `root`. What I usually do is make a same uid+gid for the MP container as the host. Makes seeding and tile manipulation easier, at least on Linux systems. 

For example in a Docker Compose file:
```
    user: ${HOST_UID_GID}
```
with 
```
export HOST_UID=$(id -u)
export HOST_GID=$(id -g)
export HOST_UID_GID="${HOST_UID}:${HOST_GID}"
```
I also maintain an MP Docker Image: https://github.com/justb4/docker-mapproxy, based on previous work by @yagajs, needs updates, but very stable for both seeding and serving, you may get/take some ideas from there. Limiting the number of `RUN` statements makes the Image much smaller. I mostly use Traefik, separate Container, not nginx as frontend...

docker/Dockerfile
weskamm: yes good point, will look into it. i think it will also reduce final image size

.github/workflows/dockerbuild.yml
```diff
@@ -0,0 +1,62 @@
...
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and push nginx image
+        uses: docker/build-push-action@v4
+        with:
+          file: ./docker/Dockerfile
+          push: true
+          build-args: |
+            NGINX=1
+            MAPPROXY_VERSION==${{ steps.meta.outputs.tags }}
```
JakobMiksch, line 39: [untested] to my understanding the target must be specified instead of the env var `NGINX`

```suggestion
          build-args: |
            MAPPROXY_VERSION==${{ steps.meta.outputs.tags }}
          target: nginx
```

.github/workflows/dockerbuild.yml
```diff
@@ -0,0 +1,62 @@
...
+            ghcr.io/mapproxy/mapproxy-nginx:${{ steps.meta.outputs.tags }}
+
+      - name: Build and push devserver image
+        uses: docker/build-push-action@v4
+        with:
+          file: ./docker/Dockerfile
+          push: true
+          build-args: |
+            DEVSERVER=1
+            MAPPROXY_VERSION==${{ steps.meta.outputs.tags }}
```
JakobMiksch, line 50: [untested] to my understanding the target must be specified instead of the env var `DEVSERVER`
```suggestion
          build-args: |
            MAPPROXY_VERSION==${{ steps.meta.outputs.tags }}
          target: development
```

doc/install_docker.rst
```diff
@@ -11,31 +29,90 @@ The community has very good experiences with the following ones:
...
 - https://hub.docker.com/r/kartoza/mapproxy (`github kartoza <https://github.com/kartoza/docker-mapproxy>`_)
 
+There are also images available that already include binaries for `MapServer` or `Mapnik`:
+
+- https://github.com/justb4/docker-mapproxy-mapserver
+- https://github.com/justb4/docker-mapproxy-mapserver-mapnik
+
 
 Quickstart
 ------------------
```
JakobMiksch, line 39:
```suggestion
Quickstart
----------
```

doc/install_docker.rst
```diff
@@ -11,31 +29,90 @@ The community has very good experiences with the following ones:
...
-In a production environment you might want to put a `nginx`_ in front of the MapProxy container, that serves as a reverse proxy.
-See the `Kartoza GitHub Repository`_ for detailed documentation and example docker-compose files. 
 
-.. _`kartoza/mapproxy`: https://hub.docker.com/r/kartoza/mapproxy
 .. _`nginx`: https://nginx.org
-.. _`GitHub Repository`: https://github.com/kartoza/docker-mapproxy
 
-There are also images available that already include binaries for `MapServer` or `Mapnik`:
+Build your own image
+-----------------------
```
JakobMiksch, line 92:
```suggestion
Build your own image
--------------------
```

.github/workflows/dockerbuild.yml
```diff
@@ -0,0 +1,62 @@
...
+
+      - name: Build and push plain image
+        uses: docker/build-push-action@v4
+        with:
+          file: ./docker/Dockerfile
+          push: true
+          build-args: |
+            MAPPROXY_VERSION==${{ steps.meta.outputs.tags }}
+          tags: |
+            ghcr.io/mapproxy/mapproxy:${{ steps.meta.outputs.tags }}
```
JakobMiksch, line 62: [untested]

```suggestion
          push: true
          build-args: |
            MAPPROXY_VERSION==${{ steps.meta.outputs.tags }}
          target: base
          tags: |
            ghcr.io/mapproxy/mapproxy:${{ steps.meta.outputs.tags }}
```

doc/install_docker.rst
```diff
@@ -1,8 +1,26 @@
 ﻿Installation via Docker
-=======================
+========================
 
+MapProxy now does have its own official docker images.
```
JakobMiksch, line 4:
```suggestion
MapProxy does have its own official docker images.
```

doc/install_docker.rst
```diff
@@ -1,8 +1,26 @@
...
+These are currently hosted on the github container registry and can be found here:
```
JakobMiksch, line 5:
```suggestion
These are currently hosted on the GitHub container registry and can be found here:
```

JakobMiksch (commented):
Great effort to create an official Docker image. Inline some review suggestions.

docker/Dockerfile
weskamm: yes but this is not the content of this PR. Feel free to open up another one for that

weskamm:
Pipeline adapted, new images will be pushed on new tags in the repo.
Container now run commands as user `mapproxy` instead of root.
Container can now be quitted using e.g. CTRL+C by using trap in bash script.

weskamm:
Thanks for reviews so far, maybe a final look @jansule ?

docker/start.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+TARGET=$1
+done=0
+trap 'done=1' TERM INT
+cd /mapproxy
+
+groupadd mapproxy && \
```
jansule, line 8: Very minor and non-blocking: This will probably lead to some warning message that the group already exists, when restarting an existing container.

jansule (approved):
Looks good to me now!

docker/start.sh
weskamm: right it does, can be fixed in follow up pr

weskamm:
Thanks for reviews!

PR #636, opened 2023-04-12
Fix installation command

zabop:
I think it was probably just a typo, fixed it now. Without `install`, installation does not work.


hwbllmnn (approved):

weskamm:
Thanks for that!

Issue #690, opened 2023-06-27
Add coverage support for caches


shimoncohen:
Mapproxy may be used to serve different kind of caches without any sources. This makes it possible to serve pre-existing data in different formats that are not served from a server.
It would be nice if it was possible to define a coverage for each cache, just as it is currently possible for sources, with [directly defining in the cache](http://mapproxy.github.io/mapproxy/sources.html#coverage) or with the [parts section](http://mapproxy.github.io/mapproxy/configuration.html) in the mapproxy.yaml.

shimoncohen:
This could be a buildup for #166.

shimoncohen:
@weskamm This issue can be closed. Thank you very much!

Issue #659, opened 2023-04-18
Add layer extent to WMTS Capabilities


asafMasa:
## Context

each layer has 2 extents
* projection extent - example for 4326 - Bounds:  180.0000, -90.0000, 180.0000, 90.0000
* layer extent - actual layer Bounds:  13.9, 41, 15, 42

## Expected Behavior
Should be able to provide layer extent for WMTS layer (use case cache type: s3)

## Actual Behavior
unable to provide layer extent for WMTS layer (use case cache type: s3)

## Possible Fix
Add support for extent for WMTS layer

## Context

wmts capabilities should provide this data for clients so they won't request irrelevant tiles 

## Your Environment

* Version used: 1.16

shimoncohen:
@weskamm I think this issue can be closed. Thank you very much!

Issue #304, opened 2017-05-04
TMS bad tileset configuration


RubMart:
Configuring a TMS layer, the configuration of the Tilesets and therefore URLs based on a Grid Mercator is not correct, there is no real level 0, and causes any level is poorly coded. Actually, it lacks the base level of the resolution of Web Mercator, since the level that is indicated like 0, is actually the level of zoom 1.

For example, the level 0 of a Grid Mercator, in units per pixel has the value 156,412 according to official sources: http://wiki.openstreetmap.org/wiki/Zoom_levels

The result obtained

<TileSet href = "http: // localhost: 8000 / tms / 1.0.0 / MICAPA / osm_grid / 0" units-per-pixel = "78271.516964" order = "0" />

This causes the actual TMS addresses to be incorrect, using a lower than real zoom level (order). It should be level 15 to which this setting corresponds, but redirects to level 14.

<TileSet href = "http: // localhost: 8000 / tms / 1.0.0 / MICAPA / osm_grid / 14" units-per-pixel = "4.77731426782" order = "14" />

Http: // localhost: 8000 / tms / 1.0.0 / MICAPA / osm_grid / 14/16079 / 20408.png

And aditionally:
1) Not configuring min_resolution in TMS layer capabilities
2) Not configuring real layer BBOX in TMS layer capabilities

## Your Environment

* Version used: 1.9.0
* Environment name and version (e.g. Python 2.7.5 with mod_wsgi 4.5.9): Python 2.7.12
* Server type and version: debug-version
* Operating System and version: ubuntu 14.04 LTE
* Configuration of my YAML file:

sources:
  MICAPA:
    type: tile
    url: http://MYCUSTOMURI
    grid: osm_grid
    min_res: 9
    transparent: true
    coverage:
      bbox: [-3.380749,40.38536570,-3.345202,40.42151]
      bbox_srs: 'EPSG:4326'

grids:
    osm_grid:
        base: GLOBAL_MERCATOR
        srs: 'EPSG:3857'
        origin: sw

olt:
The TMS specification defines that a webmercator tilegrid starts with four tiles (https://wiki.osgeo.org/wiki/Tile_Map_Service_Specification#global-mercator)

Please use ``/tiles`` instead of ``/tms`` to get the desired behavior. (See: https://mapproxy.org/docs/nightly/services.html#openlayers)

RubMart:
[...]

PR #691, opened 2023-06-27
Support cache coverage

shimoncohen:
Add coverage support for caches #690.
This also adds support for #304 and #659.


weskamm:
This sounds promising, thanks so far. Could you please add some tests?

shimoncohen:
Sure, I will work on it!

shimoncohen:
@weskamm Added tests, could you please have a look?

weskamm:
A coverage can also be a (multi)polygon, see https://mapproxy.github.io/mapproxy/coverages.html#coverages

Your code looks like that it only handles / respects extent / bbox, but not real polygons. Do you think you could extent it to also handle that? If not, it should be stated very clearly that this feature is limited to the extent of a geometry only.

shimoncohen:
Thank you very much for bringing this to my attention. I checked to make sure, and this seems to support other types of coverages as expected.
I added cases for `intersection` and `(multi)polygon` to the cache coverage tests.
Does this seem to cover this issue?

weskamm:
Well, if i define a coverage for a cache, which e.g. is a polygon of france, i would expect the cache to not have any tiles in it that are outside of france. A bbox or extent of france also covers parts of switzerland. If the implementation only handles the extent / bbox, there will be much more tiles in the cache than expected.
I am not 100% sure here but i think we currently have that issue described. Could you check for this case and add a test which checks if no tiles are chaced outside of a real polygon?

weskamm:
Could you please explain your exact use case for this pull request? After think about it for a while I am not sure if your problem could already be solved with exisiting solutions. If, for example, you have a directory with exisiting tiles which you want to filter with a coverage, you could also define the exisiting tiles as a source and use the coverage option on the source. Sorry for asking this question so late, but i am not sure if this is really necessary.

shimoncohen:
Gladly! As I mentioned in #691, from what I understand, Mapproxy's support for `sources` is only for some servers.
In our case, we want to serve data we have in `cache` types such as a `geopackage` or `s3` (In general, defining a cache with no sources as Mapproxy allows).
The goal of this PR is to add the ability to be able to define a `coverage` for each `cache` so it will be handled the same way it is currently when defined in the `sources` section.

If we would list the "sub-goals" in case you add a coverage for a `cache`:
* Extent in demo is correctly set so exact data could be displayed (not whole world extent which is default)
* Capabilities report exact BBOX for each layer and not whole world which is default
* Data out of defined coverage is not saved / displayed
* In case of `geopackage` the extent is set to be as coverage BBOX

weskamm:
Ok, i think i got a better understanding of your usecase:
You have an already existing cache filled with data. You want to filter that cache in order to not deliver the whole content, but a subset instead. You also want the capabilities to show the custom defined bbox.

Question:
As you can / need to define a grid for your cache, why dont you just define the bbox on the grid? It should solve your problems altogether, or am i missing something?

The problems i currently see with your solution:
  * You seem to only handle the bbox case, which is enough to calculate an extent / bbox for capabilities documents. But as mentioned before, coverages can be polygons. If you only need a bbox in the end, why not introduce a `bbox` parameter on the cache (if the above suggestion does not work). That would avoid confusion because people defining a polygon as coverage may expect a different behaviour than always clipping to its extent.
  * People may have an empty cache, connected with a source. And they want to seed that source. With this change, they can now configure a coverage on the cache and source. This is a somewhat conflicting situation (defining a larger area on the cache than on the source -> makes no sense, will never git filled completely; defining a smaller area on the cache than on the source -> does not make sense, use the source coverage directly). When the source has no coverage restrictions set but the cache, mapproxy will request tiles which are then not allowed to save in the cache, as the cache coverage may be smaller, which would produce undesired traffic and behaviour.
  
Please rethink the idea if a bbox is enough for your usecase and if you can just use the grid for that. If you have a valid usecase together with mapproxy configuration and data, i happily search for further solutions / ideas if you post it here.

shimoncohen:
You understand the core use case, but this is not exactly what is implemented. This addition supports all types of coverage (like with sources), not just BBOX. And the use case is also ment to support any kind of `coverage`, which is why defining a `grid` `BBOX` is not enough.

In addition, I added an example for using a `geojson` datasource as coverage to the tests. Do you see a specific case that does not work with my implementation? The only place that a `BBOX` is specifically used is in the `capabilities` and `demo` because they use `BBOX` by definition.

After your notes I also added the handling of requests outside of coverage so they won't be stored or requested from cache (or the source it is defined over). From my checks it works with the combination of coverages on `source` and `cache`.
I haven't added tests for this very important use case yet because I want to make sure you approve of the solution.

mapproxy/cache/base.py
```diff
@@ -42,6 +42,9 @@ class TileCacheBase(object):
 
     supports_timestamp = True
 
+    def __init__(self, coverage=None) -> None:
+        self.coverage = coverage
```
weskamm, line 46: Where is the coverage used in this class? I cannot see it, can you give me a point where i can see where the coverage is handled for the different caches?

shimoncohen: Sure!

The `cache` coverage is accessed in the `loader` class which calculates each caches extent in the `caches` function before return (line 1668):
https://github.com/mapproxy/mapproxy/pull/691/files#diff-7557b8794cb455eb559651f8fa4730d2a8a55d2c13bd188d730eecb4ed10e863R1668

It is also accessed in the `tile` class under cache directory in the `load_tile_coords` function in order to remove tiles that are not in the `coverage` (line 156):
https://github.com/mapproxy/mapproxy/pull/691/files#diff-cc6dc27bf169f009bbbd33027f122d17a5cac2e3f4224db2b58d3869b59673d9R156

weskamm: Thanks

weskamm:
Ok i think i am fine with the idea now, thanks so far. Please go on with the test you wanted to add.

shimoncohen:
@weskamm Added tests.

weskamm:
Thanks, this looks good. Could you finally enhance the documentation and add some information about the usecases when this should be used and the possible conflicts when misconfiguring source and chache coverage at the same time?

shimoncohen:
@weskamm Added note about using `coverages` with both `source` and it's defined `cache`.

shimoncohen:
Currently in this PR, when there is a `cache` defined over a `source`, the `coverage` used for capabilities and demo will be of the `source`.
I am wondering if this should be changed such that the `cache` coverage will be used, because it mainly defines the layer. Should this be changed?

Note: When using a `cache` with no source or `source` with no `cache` it is used correctly.

Issue #757, opened 2023-10-18
Cleanup of mbtiles seed fails in Mapproxy 1.16


alfmarcua:
In Mapproxy 1.16 there is a function remove_tile that needs now a parameter call "dimensions". That function is redefined for mbtiles but without that parameter ([https://github.com/mapproxy/mapproxy/blob/1.16.0/mapproxy/cache/mbtiles.py#L374](https://urldefense.com/v3/__https:/github.com/mapproxy/mapproxy/blob/1.16.0/mapproxy/cache/mbtiles.py*L374__;Iw!!DOxrgLBm!HIQ9pLH-JGk0jMZIHxO13nKRwwToADeRfkXeib4LJDQVtxx0Rd27AY1be2IIVLT0D0DuO94a_ochNqHG1h0-tY20KeqVShGbAq8$)) which leads to an error.

## Context

After running a cleanup to remove old tiles, the following error raised:

```
========== Cleanup tasks ==========

Start cleanup process (1 task)

  layer_seed:

    Cleaning up cache 'layer' with grid 'EPSG3857' in EPSG:3857

    Limited to coverage in: -33.75000, 27.05913, 47.81250, 80.17871 (EPSG:4326)

    Levels: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]

    Remove: tiles older than 2023-10-06 14:56:12.732830

[14:56:12]  0   0.00% -3757032.81427, 3130860.70551, 5322463.15355, 15654303.06896 (0 tiles)

Process TileCleanupWorker-1:

Process TileCleanupWorker-2:

Traceback (most recent call last):

  File "/usr/lib/python3.10/multiprocessing/process.py", line 314, in _bootstrap

    self.run()

  File "/opt/data/mapproxy/lib/python3.10/site-packages/mapproxy/seed/seeder.py", line 143, in run

    self.work_loop()

  File "/opt/data/mapproxy/lib/python3.10/site-packages/mapproxy/seed/seeder.py", line 167, in work_loop

    self.tile_mgr.remove_tile_coords(tiles)

  File "/opt/data/mapproxy/lib/python3.10/site-packages/mapproxy/cache/tile.py", line 204, in remove_tile_coords

    self.cache.remove_tiles(tiles)

  File "/opt/data/mapproxy/lib/python3.10/site-packages/mapproxy/cache/base.py", line 70, in remove_tiles

    self.remove_tile(tile, dimensions=dimensions)

[...]

PR #758, opened 2023-10-18
Quick fix for cleanup of mbtiles

alfmarcua:
Quick fix for #757 

Added the parameter dimensions to remove_tile function although it is not used.


weskamm:
Thanks for that!

PR #771, opened 2023-11-09
add tls auth

bmanojlovic:
this PR adds tls authentication to redis cache implementation, to test implementation (all test cases) now there are shell scripts in `mapproxy/test/helper_scripts/` that run three redis instances with different configurations that are possible with current code.
@shimoncohen please test in your use case if that satisfies your requirements.

### DIFF
--- a/mapproxy/cache/redis.py
+++ b/mapproxy/cache/redis.py
@@ -35,17 +35,23 @@ log = logging.getLogger(__name__)
 
 
 class RedisCache(TileCacheBase):
-    def __init__(self, host, port, prefix, ttl=0, db=0, username=None, password=None, coverage=None):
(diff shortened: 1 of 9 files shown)


weskamm:
This looks good, i just added a minor comment you may want to tackle

bmanojlovic:
where is comment? as i do not see it inline?

weskamm:
image

bmanojlovic:
this is really annoying but this is what i see
image

weskamm:
Hmm, maybe cache in your browser or some github settings? This link should also show the comment i guess:
https://github.com/mapproxy/mapproxy/pull/771/files/8bd51c11004cd3a72a29195da1336c7ced94cd83

weskamm:
Thank you, this looks good now!

PR #796, opened 2023-11-29
Fix mixed image format for file, S3 and azureblob caches

### LINKED ISSUE #605: Mixed image formats doesn't work with s3/local directories tiles cache
## Context
Mapproxy has the ability to use mixed image formats (https://mapproxy.org/docs/nightly/configuration.html#format).
It works fine when the tiles stored in a geopackage file, but doesn't work when using s3/local directory hierarchy.

## Expected Behavior
Tiles should be rendered to the client.

## Actual Behavior
Nothing is rendered. All tiles are empty.

## Possible Fix
**S3:**
When making a request ([as shown here](https://github.com/mapproxy/mapproxy/blob/master/mapproxy/cache/s3.py#L130)) the file's extension is `.mixed`. Since the original file extension isn't known, we need to consider checking for the correct extension and use it.
Worst case scenario is making two requests per tile in order to find out the extension. 

**FS**:
It's the same for FS - [this line](https://github.com/mapproxy/mapproxy/blob/master/mapproxy/cache/file.py#L114).

## Steps to Reproduce
Create a directory with mixed tiles in it, upload to it S3 / use it locally and try to configure Mapproxy to use it.

## Your Environment
* Version used: 1.15.1.post20221002
* Environment name and version:  Pyhon 3.8
* Operating System and version: Ubuntu 20.04

shimoncohen:
Fix the usage of `mixed` image format with the `File`, `S3` and `azureblob` caches.
Fixes #605.


shimoncohen:
@weskamm Could you please review this PR?

shimoncohen:
@weskamm Could you please take a look? If there are changes that need to be made I will fix them right away.

weskamm:
Looks good to me, thanks for the fixup.

simonseyock:
We are reverting this PR because it was causing problems in production - especially with existing caches. I think we need backwards compatibility and unit tests to assure that everything works correctly.

shimoncohen:
What are the problems that this causes? I will work on the issues ASAP.

simonseyock:
There was the problem that mapproxy did only return blank images for caches that were created before the update.

In the PR itself I see this problems:
1. It does not introduce a test that demonstrates the failing case
2. In the existing tests, it is ensured that every tile is stored with a `.png` ending. I do not know exactly why that is, but it feels not right to me. See https://github.com/mapproxy/mapproxy/blob/a73fe7fa696d397c14eda069e0a4340f58b8bda6/mapproxy/test/system/test_mixed_mode_format.py#L121
3. The new `tile_location` function has this code:
```
        if self.is_mixed:
            location = self._tile_location(tile, self.cache_dir, 'jpeg', create_dir=create_dir, dimensions=dimensions)
            if not os.path.exists(location):
                tile.location = None
                location = self._tile_location(tile, self.cache_dir, 'png', create_dir=create_dir, dimensions=dimensions)
```
This will never cause a file with `.jpeg` to be created. Might be intended so
4. In the `tile_location` function of the file cache there is code that does nothing. This was there before and maybe is unrelated to the problem: https://github.com/mapproxy/mapproxy/blob/a73fe7fa696d397c14eda069e0a4340f58b8bda6/mapproxy/cache/file.py#L55C13-L55C79

So, if MapProxy should supported reading caches that mix `.jpeg` and `.png` files I think it should also create `.jpeg` and `.png` files. Furthermore we need to ensure that old caches are still usable.

simonseyock:
I do not have a reproducible example, but I hope to get one.

We reverted the commit because we needed a clean working release before merging all of the linting and deprecation issues.

shimoncohen:
> There was the problem that mapproxy did only return blank images for caches that were created before the update.
> 
> In the PR itself I see this problems:
> 
> 1. It does not introduce a test that demonstrates the failing case
> 2. In the existing tests, it is ensured that every tile is stored with a `.png` ending. I do not know exactly why that is, but it feels not right to me. See https://github.com/mapproxy/mapproxy/blob/a73fe7fa696d397c14eda069e0a4340f58b8bda6/mapproxy/test/system/test_mixed_mode_format.py#L121
> 3. The new `tile_location` function has this code:
> 
> ```
>         if self.is_mixed:
>             location = self._tile_location(tile, self.cache_dir, 'jpeg', create_dir=create_dir, dimensions=dimensions)
>             if not os.path.exists(location):
>                 tile.location = None
>                 location = self._tile_location(tile, self.cache_dir, 'png', create_dir=create_dir, dimensions=dimensions)
> ```
> 
> This will never cause a file with `.jpeg` to be created. Might be intended so 4. In the `tile_location` function of the file cache there is code that does nothing. This was there before and maybe is unrelated to the problem:
> 
> https://github.com/mapproxy/mapproxy/blob/a73fe7fa696d397c14eda069e0a4340f58b8bda6/mapproxy/cache/file.py#L55
> 
> So, if MapProxy should supported reading caches that mix `.jpeg` and `.png` files I think it should also create `.jpeg` and `.png` files. Furthermore we need to ensure that old caches are still usable.

Thank you very much for your review! I will add a failing test as you requested and fix the existing tests.
As for your third point, I see that I have ignored the `create_dir` parameter that indicates wanting to store a tile insted of just getting it from the cache. I will add the use of this variable in the condition.
I agree that storing mixed tiles is what was intended.

PR #819, opened 2023-12-11
Add parameter to change background map source

alfmarcua:
Add the possibility to modify the background map shown in MapProxy, which by default takes the OpenStreetMap source.

This background map can be observed in the /demo service of MapProxy, in any of the three types of service (WMS, WMTS and TMS).

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -820,6 +820,18 @@ globals
 
 Here you can define some internals of MapProxy and default values that are used in the other configuration directives.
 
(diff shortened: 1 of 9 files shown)


weskamm:
Good idea, but documentation is missing and a hint that only XYZ Layers are allowed

alfmarcua:
Good point. Documentation added as well as an example.

weskamm:
Nice work, thank you!

PR #828, opened 2023-12-14
Configure image options for caches

hwbllmnn:
When configuring two WMS sources with transparent image options and two caches, requesting with an empty cache will result in proper merged transparent images. Once the cache contains the (properly stored transparent) tiles the merge will result in the tiles not being transparent any more as the `ImageSource` instances are created without `image_opts`.

If image options are configured on file caches they were only used to determine the file format to store the tiles. This PR changes that to also pass the image options to the `ImageSource` instances, resulting in proper behaviour when merging images.

### DIFF
--- a/doc/caches.rst
+++ b/doc/caches.rst
@@ -84,6 +84,11 @@ This is the default cache type and it uses a single file for each tile. Availabl
 
   .. versionadded:: 1.6.0
 
+``image``:
+  See :ref:`image_options` for options.
+
+  .. versionadded:: 1.16.1
+
 .. _cache_mbtiles:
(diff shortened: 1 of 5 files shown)


weskamm:
Nice! Maybe you can add tests for the new behaviour?

hwbllmnn:
> Nice! Maybe you can add tests for the new behaviour?

Funnily enough, the behaviour was already specified in the `spec.py`, so I just added a test to see that validation does not fail when caches are configured with `image`.

weskamm:
All right then, thanks for the fix!

Issue #829, opened 2023-12-19
AttributeError: type object 'TestCase' has no attribute 'failUnless' with Python 3.12


sebastic:
## Context
The testsuite fails with Python 3.12:
```
============================= test session starts ==============================
platform linux -- Python 3.12.1, pytest-7.4.3, pluggy-1.3.0
rootdir: /build/mapproxy-2.0.0+dfsg/.pybuild/cpython3_3.12_mapproxy/build
configfile: pytest.ini
collected 920 items / 82 errors

==================================== ERRORS ====================================
___________________ ERROR collecting mapproxy/test/helper.py ___________________
/usr/lib/python3/dist-packages/_pytest/runner.py:341: in from_call
    result: Optional[TResult] = func()
/usr/lib/python3/dist-packages/_pytest/runner.py:372: in <lambda>
    call = CallInfo.from_call(lambda: list(collector.collect()), "collect")
/usr/lib/python3/dist-packages/_pytest/doctest.py:567: in collect
    module = import_path(
/usr/lib/python3/dist-packages/_pytest/pathlib.py:567: in import_path
    importlib.import_module(module_name)
/usr/lib/python3.12/importlib/__init__.py:90: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
<frozen importlib._bootstrap>:1387: in _gcd_import
    ???
<frozen importlib._bootstrap>:1360: in _find_and_load
    ???
<frozen importlib._bootstrap>:1331: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:935: in _load_unlocked
    ???
<frozen importlib._bootstrap_external>:994: in exec_module
    ???
<frozen importlib._bootstrap>:488: in _call_with_frames_removed
    ???
mapproxy/test/helper.py:26: in <module>
    from mapproxy.test import mocker
mapproxy/test/mocker.py:126: in <module>
    class MockerTestCase(unittest.TestCase):
mapproxy/test/mocker.py:456: in MockerTestCase
    assertTrue = unittest.TestCase.failUnless
E   AttributeError: type object 'TestCase' has no attribute 'failUnless'
___________________ ERROR collecting mapproxy/test/mocker.py ___________________
[...]
```
The alias was removed, see: https://docs.python.org/3/whatsnew/3.12.html#unittest-testcase-removed-aliases

## Expected Behavior
No test failure

## Actual Behavior
Test failure

## Possible Fix
[...]

PR #833, opened 2023-12-20
Fix compatibility with Python 3.12.

sebastic:
Fixes: #829


weskamm:
Thanks for that, was just on it to fix it by myself :-)

PR #835, opened 2023-12-21
Docker configuration hardening 

ahennr:
This PR introduces some hardening of the Docker / NGinx configuration. In particular:

* Add an action that performs a trivy security scan on for base docker image (e.g.  `ghcr.io/mapproxy/mapproxy/mapproxy:2.0.0`) for tagged releases
* Update debian base image to `bookworm`: This will reduce the number of CVEs significantly, for example in the `nginx-image`:
   *  `docker run -it --rm ghcr.io/aquasecurity/trivy:canary image --severity CRITICAL,HIGH --report all ghcr.io/mapproxy/mapproxy/mapproxy:1.16.0`
   Result `Total: 253 (HIGH: 225, CRITICAL: 28)`
  * the updated image (local build)
   Result: `Total: 63 (HIGH: 61, CRITICAL: 2)`
* Apply some of recommendations listed in the CIS NGINX Benchmarks (see also https://www.cisecurity.org/cis-benchmarks), e.g.
  * hide server information (nginx version in headers…)
  * set proper values for timeouts

PR #841, opened 2023-12-22
Docker: introduce alpine based image for MapProxy

ahennr:
In this PR an Alpine based version of the Docker image is provided as a more lightweight alternative to the Debian based images. In comparison, the average image size could be reduced by ~30%.

image

Since no current version of NGinx is available within the official Alpine repository, The NGinx executable (only for build target `nginx`) is built and installed additionally. The version can be overwritten in the build using the parameter `NGINX_VERSION`.
The NGinx configuration already contains the changes from https://github.com/mapproxy/mapproxy/pull/835


weskamm:
What a nice pre christmas present! Thanks for that! :+1:

Issue #847, opened 2024-01-02
Skip redundant HEAD request to s3 cache


barelsh:
## Context

I set up a config with 2 layered cache and no source. (actual data is in s3, defined redis as another cache, which is used as the layer source).
While watching the logs of s3 requests we found that there are many HEAD requests that are trigered after the GET request already fetched the tile's data. It has been obsereved in case no rescale was defined, and even more significant in case rescale is defined.
here are debug logs for fetching single tile with no rescale:
```
[2023-12-31 14:04:08,311] mapproxy.config - INFO - reading: /Users/barel.shmuely/dev/libot/mapproxy/mapproxy/mapproxy/config_template/base_config/mapproxy.yaml
2023-12-31 14:04:08,486 botocore.endpoint [DEBUG] Setting s3 timeout as (60, 60)
2023-12-31 14:04:08,488 botocore.endpoint [DEBUG] Making request for OperationModel(name=HeadBucket) with params: {'url_path': '', 'query_string': {}, 'method': 'HEAD', 'headers': {'User-Agent': 'Boto3/1.34.4 md/Botocore#1.34.5 ua/2.0 os/macos#22.6.0 md/arch#arm64 lang/python#3.9.18 md/pyimpl#CPython cfg/retry-mode#legacy Botocore/1.34.5'}, 'body': b'', 'auth_path': '/raster-dev/', 'url': 'http://<my-s3-bucket>', 'context': {'client_region': 'us-east-1', 'client_config': <botocore.config.Config object at 0x130316700>, 'has_streaming_input': False, 'auth_type': 'v4', 's3_redirect': {'redirected': False, 'bucket': 'raster-dev', 'params': {'Bucket': 'raster-dev'}}, 'S3Express': {'bucket_name': 'raster-dev'}, 'signing': {'region': 'us-east-1', 'signing_name': 's3', 'disableDoubleEncoding': True}, 'endpoint_properties': {'authSchemes': [{'disableDoubleEncoding': True, 'name': 'sigv4', 'signingName': 's3', 'signingRegion': 'us-east-1'}]}}}
[...]

PR #848, opened 2024-01-02
Skipping redundant read for cache

barelsh:
fix issue #847

The cache requests are making redundant HEAD requests. In the following cases there is no need to read from cache:
1. in case no source is defined, no rescale_tiles defined. the tile is already read into memory, or missing from the cache.
2. in case no source is defined,  rescale_tiles is defined. in this case no need to check if tile exist in cache, we can check `tile.is_missing()`

I verified that tests run successfully, especially the following test classes which are related to the changes:
```
TestTileManagerCacheGeojsonCoverage
TestTileManagerCacheBboxCoverage
TestTileManagerRescaleTiles
```


barelsh:
@weskamm I'll appreciate your review on this one

simonseyock:
That looks quite sensible, thanks for this. Is it possible to include a test for this?

mapproxy/cache/tile.py
```diff
@@ -181,8 +181,18 @@ def _load_tile_coords(self, tiles, dimensions=None, with_metadata=False,
...
+        if (self.rescale_tiles == 0 and cache_only):
+            return tiles
+
         for tile in tiles:
-            if tile.coord is not None and not self.is_cached(tile, dimensions=dimensions):
+            # in cache_only mode, we already fetched the tile from cache
+            if tile.coord is not None and (
+                (cache_only and tile.is_missing()) or
+                (not cache_only and not self.is_cached(tile, dimensions=dimensions))
+            ):
```
simonseyock, line 195: It might be nicer (i.e. more readable) to move this check into a function.

simonseyock:
Thanks for the refactoring. The code changes look good to me. The test is not as verbose as I wished it to be, but I think it is ok :)

Thanks for your work!

simonseyock (approved):

PR #861, opened 2024-01-09
remove py2 compat

simonseyock:
This removes the python 2 compatibility functions.


mapproxy/config/config.py
```diff
@@ -194,7 +196,7 @@ def load_config(config, config_file=None, config_dict=None, clear_existing=False
     defaults = _to_options_map(config_dict)
 
     if defaults:
-        for key, value in iteritems(defaults):
+        for key, value in defaults.items(): # todo
```
weskamm, line 199: why todo?

mapproxy/service/wmts.py
```diff
@@ -332,7 +331,7 @@ class WMTSTileLayer(object):
     def __init__(self, layers):
         self.grids = [lyr.grid for lyr in layers.values()]
         self.layers = layers
-        self._layer = layers[next(iterkeys(layers))]
+        self._layer = layers[layers.keys()[0]]
```
weskamm, line 334: are you sure that this behaves the same? (next vs [0])

mapproxy/test/http.py
```diff
@@ -24,18 +24,10 @@
...
 from contextlib import contextmanager
+from urllib.parse import parse_qsl, urlsplit
+
 from mapproxy.util.py import reraise
-from mapproxy.compat import iteritems, PY2
-from mapproxy.compat.modules import urlparse, parse_qsl
-if PY2:
-    from cStringIO import StringIO
-else:
-    from io import StringIO
```
weskamm, line 33: was this unused?

mapproxy/util/ext/odict.py
```diff
@@ -107,8 +107,8 @@ def __setitem__(self, key, value):
     :license: modified BSD license.
 """
 from __future__ import absolute_import
-from mapproxy.compat import iteritems
-from mapproxy.compat.itertools import izip, imap
+
+from past.builtins import cmp
```
weskamm, line 111: unused import?

weskamm (commented):
Found some minors by scrolling through the changes. Looking good otherwise!

mapproxy/config/config.py
simonseyock: I was not sure if the code was correct before. But i will leave it as it was before.

mapproxy/test/http.py
simonseyock: yes

mapproxy/util/ext/odict.py
simonseyock: is it? cmp was a builtin in the past and now we need to import it if we want to use it. I did not look into how to replace it as it was only this one place.

simonseyock: usage is here: https://github.com/mapproxy/mapproxy/pull/861/files/4337202090ec36b353dbb790b325b70b1541698a#diff-ddc40d8d20c2770cab74d4d3847fc1a8f178b463da616e6e13eac8041b0456f3L237

mapproxy/config/config.py
simonseyock:
```suggestion
        for key, value in defaults.items():
```

simonseyock:
Thanks for reviewing!

Issue #538, opened 2021-09-06
mapproxy 1.13.1 and mapnik 4.0.0 projection inicialization


mar0321:
Hi i know there was change from proj4 to newer version so did mapnik.
But i solve that in mapnik so it can read new proj library but mapproxy when it's inicialized with mapnik it throws:
SourceError("failed to initialize projection with: '+init=epsg:3857'")

i have pyproj newest version. Any help how to solve this?

sebastic:
[`mapproxy/source/mapnik.py`](https://github.com/mapproxy/mapproxy/blob/master/mapproxy/source/mapnik.py#L123) needs to be changed to use the new syntax like [python-mapnik](https://github.com/mapnik/python-mapnik/commit/ca66af65204d68a5496a94d36d69bf61144daf3b) e.g.:
```diff
--- a/mapproxy/source/mapnik.py
+++ b/mapproxy/source/mapnik.py
@@ -120,7 +120,7 @@ class MapnikSource(MapLayer):
 
         m = self.map_obj(mapfile)
         m.resize(query.size[0], query.size[1])
-        m.srs = '+init=%s' % str(query.srs.srs_code.lower())
+        m.srs = str(query.srs.srs_code.lower())
         envelope = mapnik.Box2d(*query.bbox)
         m.zoom_to_box(envelope)
         data = None
```
Note that Mapnik 4.0.0 is not released yet, PROJ 6+ changes are only available in the [mapnik](https://github.com/mapnik/mapnik/pull/4202) & [python-mapnik](https://github.com/mapnik/python-mapnik/commits/proj6) git repos.

justb4:
Suggestion: maybe it is possible to support both `srs`-assignment syntaxes? Maybe on Mapnik version-check or sometimes from an Exception, like we do for `lxml`?

mar0321:
> [`mapproxy/source/mapnik.py`](https://github.com/mapproxy/mapproxy/blob/master/mapproxy/source/mapnik.py#L123) needs to be changed to use the new syntax like [python-mapnik](https://github.com/mapnik/python-mapnik/commit/ca66af65204d68a5496a94d36d69bf61144daf3b) e.g.:
> 
> ```diff
> --- a/mapproxy/source/mapnik.py
> +++ b/mapproxy/source/mapnik.py
> @@ -120,7 +120,7 @@ class MapnikSource(MapLayer):
>  
>          m = self.map_obj(mapfile)
[...]

PR #886, opened 2024-02-07
Fixed deprecated proj4 interface usage in source/mapnik.py

glebpinigin:
Mapnik source on-the-fly reprojection used deprecated proj4 API for assigning mapnik.Map object's projection.

This behavior and the way to fix it is described in issue #538 

Debian package maintainers apply this fix with [one of patches](https://salsa.debian.org/debian-gis-team/mapproxy/-/blob/master/debian/patches/mapnik.patch?ref_type=heads) so that debian-based package works properly and does not fail to render mapnik sources. However it is not the case for installations based on source code (e.g. via pip)

### DIFF
--- a/mapproxy/source/mapnik.py
+++ b/mapproxy/source/mapnik.py
@@ -231,7 +231,7 @@ class MapnikSource(MapLayer):
 
         m = self.map_obj(mapfile)
         m.resize(query.size[0], query.size[1])
-        m.srs = '+init=%s' % str(query.srs.srs_code.lower())
+        m.srs = str(query.srs.srs_code.lower())
         envelope = mapnik.Box2d(*query.bbox)
         m.zoom_to_box(envelope)
         data = None


glebpinigin:
This fix may lead to some backwards compatibility issues with outdated versions of mapnik and proj. I am not sure if it is an issue, because most popular distributions ship mapnik versions that support newer proj API thus incompatible with current mapproxy source code.

However, if we don't want to deal with this compatibility break there is a way to generate valid proj string from EPSG code with pyproj and using it for mapnik configuration

b-r-u:
What about something like this? It needs an import from pyproj that I see nowhere else, but that library seems to be a requirement for mapproxy.

```python
from pyproj import PROJ_VERSION
#....
# Check major proj version and adjust srs definition accordingly
if PROJ_VERSION[0] >= 6:
    m.srs = str(query.srs.srs_code.lower())
else:
    m.srs = '+init=%s' % str(query.srs.srs_code.lower())
```

LouisPauchet:
This would be a very handy fix to be included in the project avoiding to require a manual patch of if.  :-)

PR #909, opened 2024-03-14
Add requires that are needed for the dev image to run

simonseyock:
I do not know exactly if this is the desired way, but this seems to be needed in order to run the `-dev` image. The `-nginx` image seems to be a little more robust, but in my opinion this should not harm and will also get rid of the warning when starting the container (`Found libproj >=5. Using this library without pyproj is deprecated and not fully supported. Please install pyproj >= 2.`)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,6 +5,8 @@ import pkg_resources
 
 install_requires = [
     'PyYAML>=3.0',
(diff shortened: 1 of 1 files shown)


weskamm:
Thanks for that!

PR #925, opened 2024-04-19
Simplify check for dimensions support

mulimoen:
This opens up for creating or modifying other cache types to also support dimensions.


simonseyock (approved):
This looks good to me. Thanks for the good work!

PR #931, opened 2024-05-24
Fix broken TMS/WMTS demo templates HTML

justb4:
The TMS demo was broken after the upgrade to OpenLayers 7. The template HTML assumes TMS access as an XYZ OL Class with swapped Y `{-y}`, tile coord. This created at least two problems that are fixed in this PR:

* the TMS URL path  should contain the `/1.0.0/` substring
* support for a custom Grid `ol.tilegrid.TileGrid` was lacking

Edit: WMTS demo template uses `ol.tilegrid.TileGrid`, but its constructor lacks a `tileSize` attribute (e.g.  custom for HQ/Retina 512x512 tiles).

These issues appeared when upgrading the [justb4/mapproxy Docker image](https://github.com/justb4/docker-mapproxy) (to MP 2.0.2). For now the Docker image is patched during builds using [this patch file](https://github.com/justb4/docker-mapproxy/blob/master/patches/03-tms-demo/tms-demo.patch). 

There may be additional improvements like explicit Origin (always confusing with custom grids and TMS), grid custom tile_size like for HQ/Retina tiles in both TMS and WMTS demo templates. Tile size from grid will be included in this PR.
IMHO `origin` is implicit (`sw`/`ll`) for TMS and already specified (`nw`/`ul`) for WMTS.

### DIFF
--- a/mapproxy/service/templates/demo/tms_demo.html
+++ b/mapproxy/service/templates/demo/tms_demo.html
@@ -1,7 +1,7 @@
 {{py:
 from html import escape
 import textwrap
-
+import json
 wrapper = textwrap.TextWrapper(replace_whitespace=False, width=90,
                                break_long_words=False)
 
@@ -23,7 +23,7 @@ jscript_functions=None
         const transparent = {{if format == 'png'}}true{{else}}false{{endif}};
         const format = "{{format}}";
         const srs = "{{srs}}";
-        const tileSize = "{{layer.grid.tile_size}}";  // TODO allow custom e.g. Retina tilesizes
+        const tileSize = {{json.dumps(layer.grid.tile_size)}};
         const extent = [{{', '.join(str(r) for r in layer.bbox)}}];
         const grid_extent = [{{', '.join(str(r) for r in layer.grid.bbox)}}];
         const resolutions = [{{', '.join(str(r) for r in resolutions)}}];
@@ -46,7 +46,7 @@ jscript_functions=None
             projection: "{{srs}}",
             maxResolution: {{resolutions[0]}},
             tileGrid: new  ol.tilegrid.TileGrid({
-               tileSize: 256,
+               tileSize: tileSize,
                resolutions: resolutions,
                extent: grid_extent
             }),
--- a/mapproxy/service/templates/demo/tms_demo.html
+++ b/mapproxy/service/templates/demo/tms_demo.html
@@ -20,9 +20,13 @@ jscript_functions=None
 <script type="text/javascript">
     async function init() {
 
-        const transparent = "{{if format == 'png'}}true{{endif}}";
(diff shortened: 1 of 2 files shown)


mapproxy/service/templates/demo/tms_demo.html
```diff
@@ -35,11 +39,17 @@
...
         const source = new ol.source.XYZ({
-            url: '../tms/{{"/".join(layer.md["name_path"])}}/{z}/{x}/{-y}.png',
-            opaque: transparent === "true" ? false : true,
+            url: '../tms/1.0.0/{{"/".join(layer.md["name_path"])}}/{z}/{x}/{-y}.' + format,
+            opaque: transparent,
             projection: "{{srs}}",
-            maxResolution: {{resolutions[0]}}
+            maxResolution: {{resolutions[0]}},
+            tileGrid: new  ol.tilegrid.TileGrid({
+               tileSize: 256,
```
walkermatt, line 49: @justb4 should this be using the `tileSize` variable defined on line 26?

walkermatt: Actually I think line 26 should be along these lines (assuming `json` has been imported):
```
        const tileSize = {{json.dumps(layer.grid.tile_size)}};  // TODO allow custom e.g. Retina tilesizes
```
While testing the value of `tileSize` was `"(256, 256)"` for the demo OSM layer.

justb4: @walkermatt yes,  `tileSize` (and maybe `origin`) was not present even in the pre-OL7 version. Had to figure out how to convert from what looks like a stringified Python `tuple` to a JS `array`, but you nailed it, thanks! Edit upcoming.

justb4: and the [WMTS demo HTML template](https://github.com/mapproxy/mapproxy/blob/master/mapproxy/service/templates/demo/wmts_demo.html#L55) also lacks the tile size, will add there as well and slightly change PR title/desc. (found that while working on HQ 512x512 tiles).

justb4: Also fixed tile sizes in WMTS and TMS. 

Only remaining item is `origin`: as TMS is emulated using the class `ol.source.XYZ` with flipped `y` `{-y}`, I wondered if `origin` should be set explicit similar to WMTS, like: ` origin: ol.extent.getBottomLeft(grid_extent)`  But that had reverse effects. IMHO origin of TMS is implicit (lower-left) and MapProxy assumes that, so independent how the internal tile-grid is configured  (`sw` or `nw`), the proper tiles are fetched.  I keep it this way (no explicit `origin` for TMS).

So al in all, think good to go, pending reviews.

hwbllmnn (approved):

weskamm:
Thanks for your contribution!

Issue #928, opened 2024-05-15
'bulk_meta_tiles' does not work for 'sqlite' caches


xdml:
## Context

When enabled `bulk_meta_tiles` mapproxy resulted in error for `sqlite` caches. The error follows:

``` python
TypeError: MBTilesLevelCache.store_tiles() missing 1 required positional argument: 'dimensions'
```

## Expected Behavior

Just works with enabled `bulk_meta_tiles` for `sqlite` caches.

## Actual Behavior

Cannot use `bulk_meta_tiles`.

## Possible Fix

Check `_create_bulk_meta_tile(self, meta_tile)` in [mapproxy](https://github.com/mapproxy/mapproxy/tree/master)/[mapproxy](https://github.com/mapproxy/mapproxy/tree/master/mapproxy)/[cache](https://github.com/mapproxy/mapproxy/tree/master/mapproxy/cache)/[tile.py](https://github.com/mapproxy/mapproxy/blob/master/mapproxy/cache/tile.py) and possibly fix incorrect calling of:

```python
self.cache.store_tiles([t for t in tiles if t.cacheable])
```

as `dimensions` argument is missing.

Not sure, if it can be fixed by:

```python
self.cache.store_tiles([t for t in tiles if t.cacheable], dimensions=self.dimensions)
```

as I do not have deep knowledge about `mapproxy` insides.

## Steps to Reproduce

1. Configure tiled source `cache` with `sqlite` storage type
2. Configure `meta_size: [5, 5]` in `globals: cache:`
3. Configure `bulk_meta_tiles: true` in `globals: cache:`
4. Try to access any configured service

## Context

Want to use `sqlite` storage type, as I have issues with `file` storage type (too many small files, try to reduce inode usage).

## Your Environment

* Version used: 1.15.1 (debian package)
* Environment name and version (e.g. Python 2.7.5 with mod_wsgi 4.5.9): Python 3.11.2 with gunicorn 20.1.0 (both debian packages)
* Server type and version: LXC container
* Operating System and version: Debian GNU/Linux 12 (bookworm)

Even if I do not use `mapproxy` installed from `pip`, it looks this issue still exists in the code.

xdml:
Seeding not working as well if using `sqlite` cache.

```python
TypeError: MBTilesLevelCache.load_tile_metadata() missing 1 required positional argument: 'dimensions'
```

simonseyock:
thanks for reporting and fixing this @xdml.

PR #933, opened 2024-06-02
Added missing `dimensions` keyword argument to `load_tile_metadata()` and `store_tiles()` calls

xdml:
#928


simonseyock (approved):
Thanks for this fix @xdml

Issue #935, opened 2024-06-11
GeoPackage Cache: occasional locking errors during seeding


justb4:
NB: I will provide a PR for this. Edit: the same problem may affect [MBTiles caches](https://github.com/mapproxy/mapproxy/blob/master/mapproxy/cache/mbtiles.py).

## Context
Seeding a GeoPackage cache from a Mapnik+PostGIS source. First using single cache file, later leveled caches (one per zoomlevel). 
## Expected Behavior
Log output should not contain locking exception stack traces.

## Actual Behavior
See occasional exceptions like these:

```
[17:20:02]  0   0.00% 303982.99573, 6522640.17428, 844765.40321, 7139033.46273 (0 tiles)
[17:20:32] 16   2.83% 310640.08295, 7125154.02863, 313086.06786, 7127600.01354 (7672 tiles)
[17:21:02] 17   3.59% 306971.10559, 7109255.12675, 308194.09805, 7110478.11920 (14796 tiles)
[17:21:32] 17   4.21% 308194.09805, 7095802.20977, 309417.09050, 7097025.20222 (22936 tiles)

Process TileSeedWorker-1:
Traceback (most recent call last):
  File "/usr/lib/python3.7/multiprocessing/process.py", line 297, in _bootstrap
    self.run()
  File "/usr/local/lib/python3.7/dist-packages/mapproxy/seed/seeder.py", line 143, in run
    self.work_loop()
  File "/usr/local/lib/python3.7/dist-packages/mapproxy/seed/seeder.py", line 158, in work_loop
    exceptions=(SourceError, IOError), ignore_exceptions=(LockTimeout, ))
  File "/usr/local/lib/python3.7/dist-packages/mapproxy/seed/util.py", line 189, in exp_backoff
    result = func(*args, **kw)
  File "/usr/local/lib/python3.7/dist-packages/mapproxy/cache/tile.py", line 141, in load_tile_coords
    rescale_till_zoom=rescale_till_zoom, rescaled_tiles={},
  File "/usr/local/lib/python3.7/dist-packages/mapproxy/cache/tile.py", line 162, in _load_tile_coords
    self.cache.load_tiles(tiles, with_metadata)
  File "/usr/local/lib/python3.7/dist-packages/mapproxy/cache/geopackage.py", line 569, in load_tiles
    return self._get_level(level).load_tiles(tiles, with_metadata=with_metadata, dimensions=dimensions)
[...]

PR #936, opened 2024-06-11
#935 GeoPackage Cache: add timeout param to SQLite connect() calls for "verify" methods

justb4:
Fix for issue #935: by adding a `timeout` parameter (provided with constructor, default 30ms) locking errors are avoided,  at least brought down to a minimum. In our seeding sessions, multiple, even for HQ/Retina 512x512 tiles never observed since this fix.


simonseyock (approved):
It makes sense for me to apply this parameter to every connect call instead of just the ones where `db()` is called. Thanks for the contribution!

Issue #937, opened 2024-06-13
Dockerfile CMD instead of ENTRYPOINT


simonseyock:
It would be nicer if the Dockerfile used a CMD instead of an ENTRYPOINT command to start it's service.

It normally is good practice to do setup work in the entrypoint script and at the end just run the cmd, which will be passed as an argument with `exec "$@"`.

This would make it nicer to use the image for seeding with:
```
docker run mapproxy mapproxy-seed -f mapproxy.yaml -s seed.yaml
```

instead of
```
docker run --entrypoint /usr/local/bin/mapproxy-seed mapproxy -f mapproxy.yaml -s seed.yaml
```

which misses some setup work done in the entrypoint script.

weskamm:
Sounds good, please just go ahead

PR #938, opened 2024-06-13
Dockerfile better entrypoint

simonseyock:
This is a draft for solving #937

It is not fully tested.


simonseyock:
This is now fully tested. The CI process is not yet adapted.

simonseyock:
I updated the dockerbuild github action. I removed the workflow_dispatch trigger because:
a) it made the action way more complicated (double build definitions for every flavor)
b) a dockerfile should *always* correspond to a git tag. so if a tag should be republished it needs to be deleted and recreated

Dockerfile
```diff
@@ -0,0 +1,104 @@
...
+
+
+FROM base-libs AS builder
+
+RUN mkdir /mapproxy
+
+WORKDIR /mapproxy
+
+COPY setup.py README.md CHANGES.txt MANIFEST.in ./
+COPY mapproxy mapproxy
```
simonseyock, line 27: These are the only files that will be included in the wheel. Maybe we need some more?

weskamm:
> I updated the dockerbuild github action. I removed the workflow_dispatch trigger because: ...
As long as we dont need to publish older images again (which i currently see no reason for), this should be fine

Dockerfile
weskamm: License at least

weskamm (approved):
Thanks for your work! Lets see if tests from Hannes succeed before merging, but you get my thumbs up!

hblitza:
Thanks @simonseyock. I tested it locally - successfully for dev,plain,nginx both alpine and debian.

hblitza (approved):

PR #939, opened 2024-06-18
build: require boto library

simonseyock:


weskamm:
We have to decide if we want to include all optional requirements, which are normally installed by hand as described on the caches page here:
https://mapproxy.github.io/mapproxy/latest/caches.html

At least for the docker images this would be nice to have

simonseyock:
I will include it here: https://github.com/mapproxy/mapproxy/pull/938

PR #940, opened 2024-07-03
fix: show all schema errors and update documentation regarding caches without source

simonseyock:
adds jsonschema and importlib_resources dependencies

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,7 +7,9 @@ from setuptools import setup, find_packages
 install_requires = [
     'PyYAML>=3.0',
     'future',
-    'pyproj>=2'
+    'pyproj>=2',
+    'jsonschema',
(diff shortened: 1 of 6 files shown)


hblitza (approved):

setup.py
```diff
@@ -7,7 +7,9 @@
 install_requires = [
     'PyYAML>=3.0',
     'future',
-    'pyproj>=2'
+    'pyproj>=2',
+    'jsonschema',
```
weskamm, line 11: Do we need documentation for this, as e.g. seen here: https://mapproxy.github.io/mapproxy/latest/install.html ?

simonseyock: We could document every dependency there ... I don't know if that is really necessary?

PR #944, opened 2024-07-04
Release workflow

simonseyock:
This was tested on the PyPI test instance: https://test.pypi.org/project/MapProxyTest

The corresponding workflow run is this one: https://github.com/simonseyock/mapproxy/actions/runs/9797894513/job/27055393588

**Before we configure this we should consider who should be able to push tags to this repository which would trigger the release.**

Here are some security considerations: https://docs.pypi.org/trusted-publishers/security-model/

To be able to publish on PyPI for real this workflow needs to be added as trusted publisher for MapProxy. It is explained here:
https://docs.pypi.org/trusted-publishers/adding-a-publisher/

with:
* owner: mapproxy
* repository: mapproxy
* workflow name: release.yml
* environment name: pypi

After this any tag that gets pushed will trigger a release on PyPI


simonseyock:
This also removes appveyor as this is not used anywhere as I can see (can be readded)
It removes the setup.cfg, which also not seems to be needed. The unversal tag for bdist is not needed, as we only publish for python 3, this also removes `py3` from the wheel name.
The `release.py` file is also not used and therefore removed.

weskamm:
Thanks for this!

PR #945, opened 2024-07-09
feat: include WMS legend url in WMTS capabilities

simonseyock:
If WMTS and WMS are enabled and a layer has configured a `legendurl`, the WMTS Capabilities will advertise the WMS `GetLegendGraphic` URL as a `LegendURL`.

### DIFF
--- a/mapproxy/config/loader.py
+++ b/mapproxy/config/loader.py
@@ -1957,6 +1957,7 @@ class LayerConfiguration(ConfigurationBase):
                         md=md,
                         tile_manager=cache_source,
                         dimensions=dimensions,
+                        legend='legendurl' in self.conf and 'wms' in self.context.services.conf
                     )
                 )
 
--- a/mapproxy/config/loader.py
+++ b/mapproxy/config/loader.py
@@ -1950,6 +1950,12 @@ class LayerConfiguration(ConfigurationBase):
                 md['format'] = self.context.caches[cache_name].image_opts().format
                 md['cache_name'] = cache_name
                 md['extent'] = extent
+                legend_version = None
+                if 'legendurl' in self.conf:
+                    wms_conf = self.context.services.conf.get('wms', None)
+                    # GetLegendGraphic with legendurl does not seem to work in non 1.3.0 versions
+                    if wms_conf is not None and '1.3.0' in wms_conf.get('versions', ['1.3.0']):
+                        legend_version = '1.3.0'
                 tile_layers.append(
                     TileLayer(
                         self.conf['name'], self.conf['title'],
@@ -1957,7 +1963,7 @@ class LayerConfiguration(ConfigurationBase):
                         md=md,
                         tile_manager=cache_source,
                         dimensions=dimensions,
-                        legend='legendurl' in self.conf and 'wms' in self.context.services.conf
+                        legend_version=legend_version
                     )
                 )
 
--- a/mapproxy/config/loader.py
+++ b/mapproxy/config/loader.py
@@ -31,6 +31,7 @@ import hashlib
 import warnings
 from copy import deepcopy, copy
 from functools import partial
+from packaging.version import Version
 
 import logging
 from urllib.parse import urlparse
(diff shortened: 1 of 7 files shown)


mapproxy/config/loader.py
```diff
@@ -1951,13 +1951,20 @@ def tile_layers(self, grid_name_as_path=False):
                 md['format'] = self.context.caches[cache_name].image_opts().format
                 md['cache_name'] = cache_name
                 md['extent'] = extent
+                legend_version = None
+                if 'legendurl' in self.conf:
+                    wms_conf = self.context.services.conf.get('wms', None)
```
jansule, line 1956: minor: you can omit the `None` argument, as this is the default

mapproxy/service/tile.py
```diff
@@ -204,7 +204,7 @@ def _render_root_resource_template(self, service):
 
 
 class TileLayer(object):
-    def __init__(self, name, title, md, tile_manager, info_sources=[], dimensions=None):
+    def __init__(self, name, title, md, tile_manager, info_sources=[], dimensions=None, legend_version=None):
```
jansule, line 207: Not directly related but it is not recommended to use `[]` as default values since the default `[]` will be a single instance for all calls. This might lead to call by reference errors. Something like this is more stable:

```
def __init__(self, name, title, md, tile_manager, info_sources=None, dimensions=None, legend_version=None):
  if info_sources is None:
    info_sources = []
```

jansule (approved):
Very nice! Just some minor remarks

mapproxy/service/templates/wmts100capabilities.xml
```diff
@@ -97,6 +97,12 @@
       <ows:Identifier>{{layer.name}}</ows:Identifier>
       <Style>
         <ows:Identifier>default</ows:Identifier>
+        {{if layer.legend_version}}
+        <LegendURL
+          format="image/png"
+          xlink:href="{{service.url}}?service=WMS&amp;request=GetLegendGraphic&amp;version={{layer.legend_version}}&amp;format=image%2Fpng&amp;layer={{layer.name}}"
```
marcjansen, line 103: Can we safely assume that `format=image/png` is supported?

mapproxy/config/loader.py
```diff
@@ -1951,13 +1951,20 @@ def tile_layers(self, grid_name_as_path=False):
...
+                    # GetLegendGraphic with legendurl does not seem to work in non 1.3.0 versions
```
weskamm, line 1957: Maybe add this to the documentation of the legendUrl parameter.

weskamm (approved):
Smart move, looking good!

mapproxy/service/templates/wmts100capabilities.xml
simonseyock: Yes, MapProxy automatically converts image formats. We could also advertise all other possible image formats if we want to.

mapproxy/config/loader.py
simonseyock: This is a bug in MapProxy and should be changed as soon as the bug is addressed. Alternatively we could add code here that chooses the highest version number so it is assuming a correctly working `GetLegendGraphic` and works as soon as the bug is fixed.

I opened this issue: https://github.com/mapproxy/mapproxy/issues/946

mapproxy/config/loader.py
simonseyock: thanks for the hint

mapproxy/service/templates/wmts100capabilities.xml
simonseyock: I will leave it at png for now, if other formats are needed we can change accordingly

mapproxy/config/loader.py
simonseyock: I changed it so it should be fixed as soon as the issue is addressed

mapproxy/service/tile.py
simonseyock: good information. this should be addressed in the whole code base I think. Maybe we do an issue for this one and solve it separately.

weskamm (approved):

PR #951, opened 2024-07-16
Test splitted build

simonseyock:
This splits off the docker build into multiple concurrently running jobs, one for ubuntu and one for alpine

This is a test run: https://github.com/simonseyock/mapproxy/actions/runs/9956997508 (trivy fails because the images are not uploaded)


simonseyock:
it also fixes the broken alpine build

weskamm (approved):

PR #956, opened 2024-07-17
feat: add plugin interface to intercept any requests before they are handled

simonseyock:
title says it all.

### DIFF
--- a/mapproxy/wsgiapp.py
+++ b/mapproxy/wsgiapp.py
@@ -111,6 +111,13 @@ def wrap_wsgi_debug(app, conf):
     return app
 
 
+request_interceptors = set()
+
+
+def register_request_interceptor(interceptor):
+    request_interceptors.add(interceptor)
+
+
 class MapProxyApp(object):
     """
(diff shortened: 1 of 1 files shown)


weskamm:
This should be already possible as described here i guess:
https://mapproxy.github.io/mapproxy/latest/auth.html#wsgi-filter-middleware
?

simonseyock:
It might be possible, but a plugin is easy to install for the end customer and this extension opens up new ways to write plugins for MapProxy. So I would prefer doing it this way.

But one could argue that it should not be possible for a plugin to work at such a crucial point, potentially breaking anything. I think it is ok, though. The plugin could break the main software in a thousand other ways as well.

weskamm:
The difficulty of using plugins is more or less the same as using a middleware, both do not require much work.
And intercepting requests is proven to work in the middleware, we are using it for several years.
But i dont mind having the additional option to intercept in plugins, so thumbs up from my side.

weskamm (approved):
Can you add doc / example in a follow up PR?

Issue #955, opened 2024-07-17
using importlib.resources instead of importlib_resources


landryb:
## Context

packaging 2.1.1 on OpenBSD, i find the new `importlib_resources` requirement added in #863. we don't have that python library packaged, and from my understanding this is only for python 3.8.

for the other newer versions of python, apparently using `importlib.resources` instead works out of the box  (i've locally tested it and `mapproxy-util` starts), so is it possible to make the requirement conditional to the python version, and adapt the corresponding codepaths to use the funcs from the stdlib ?

## Expected Behavior
no need for importlib_resources on python 3.11

## Actual Behavior
useless requirement

## Possible Fix
make it conditional

## Steps to Reproduce

1.
2.
3.
4.

## Your Environment

* Version used: 2.11
* Environment name and version (e.g. Python 2.7.5 with mod_wsgi 4.5.9): python 3.11
* Server type and version:
* Operating System and version: OpenBSD-current

landryb:
a potential way to deal with this, courtesy of @sthen: https://pbot.rmdir.de/pk_c9u8AVaq2tQ2v1t0_fA

PR #957, opened 2024-07-18
only require importlib_resources on python 3.8 (#955)

landryb:
actually tested this time, `mapproxy-util` starts fine here on 3.11

### DIFF
--- a/mapproxy/image/message.py
+++ b/mapproxy/image/message.py
@@ -17,7 +17,10 @@ from __future__ import division
 
 import logging
 import os
-import importlib_resources
+try:
+    from importlib import resources as importlib_resources
+except ImportError:
+    import importlib_resources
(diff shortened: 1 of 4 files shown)


landryb:
the failing tests on 3.8 dont seem related to the PR changes..

simonseyock:
Thanks for your contribution! I restarted the test, they sometimes have hiccups

simonseyock:
The failing tests are saying now:
```
AttributeError: module 'importlib.resources' has no attribute 'files'
```
so it seems to be related.

I think the problem is that `from importlib import resources` is available on python 3.8, but is not providing all the needed functions. A solution for this to switch around your check: always try to import `importlib_resources` and use system library only if `importlib_resources` is not installed.

PR #958, opened 2024-07-22
Make WMTS LegendURL configurable

simonseyock:
Adds two new config parameters for layers: `wmts_kvp_legendurl` and `wmts_rest_legendurl`. If those are set, MapProxy will not try to generate a LegendURL on its own (if `legendurl` was provided), but will instead use this one (no matter if the `legendurl` was provided or not).

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -395,16 +395,17 @@ Each dimension is another dictionary with a list of ``values`` and an optional `
 ``wmts_kvp_legendurl`` and ``wmts_rest_legendurl``
 """"""""""""""""""""""""""""""""""""""""""""""""""
 
-It is possible which URLS the WMTS will advertise as LegendURLs through its capabilities. These will be inserted as a
-``LegendURL`` tag into the capabilities document. These URLs support two template variables ``{base_url}`` which is the
-base url of MapProxy and ``{layer_name}`` which is the name of the layer.
+It is possible to configure which URLS the WMTS will advertise as LegendURLs through its capabilities. These will be
+inserted as a ``LegendURL`` tag into the capabilities document. These URLs support two template variables ``{base_url}``
+which is the base url of MapProxy and ``{layer_name}`` which is the name of the layer.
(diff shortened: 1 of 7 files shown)


doc/configuration.rst
```diff
@@ -392,6 +392,20 @@ Each dimension is another dictionary with a list of ``values`` and an optional `
             - 1000
             - 3000
 
+``wmts_kvp_legendurl`` and ``wmts_rest_legendurl``
+""""""""""""""""""""""""""""""""""""""""""""""""""
+
+It is possible which URLS the WMTS will advertise as LegendURLs through its capabilities. These will be inserted as a
```
weskamm, line 398:
```suggestion
It is possible to configure which URLS the WMTS will advertise as LegendURLs through its capabilities. These will be inserted as a
```

doc/configuration.rst
```diff
@@ -392,6 +392,20 @@ Each dimension is another dictionary with a list of ``values`` and an optional `
...
+``LegendURL`` tag into the capabilities document. These URLs support two template variables ``{base_url}`` which is the
+base url of MapProxy and ``{layer_name}`` which is the name of the layer.
+
+.. code-block:: yaml
+
+  layers:
+    - name: legend_layer
+      tile_sources: [cache]
+      wmts_rest_legendurl: "{base_url}/legend/{layer_name}.png" # would become "http://127.0.0.1/legend/osm.png" on localhost
```
weskamm, line 407:
```suggestion
      wmts_rest_legendurl: "{base_url}/legend/{layer_name}.png" # would become "http://127.0.0.1/legend/legend_layer.png" on localhost
```

weskamm (commented):
Can you add an configuration / outcome example for wmts_kvp_legendurl ?

doc/configuration.rst
simonseyock: nice catch

weskamm (approved):

PR #964, opened 2024-07-24
ci: fix gh-pages workflow

simonseyock:
And still something in the gh-pages workflow was broken.


weskamm:
We will see if it works out :-)

PR #967, opened 2024-08-21
Add proper XML Exception when service parameter disabled / invalid or missing

weskamm:
This adds OWS Exceptions to requests that are invalid:
- If the `service` parameter is missing, an Exception like this is thrown (with status code 400):
    ```
    <ows:ExceptionReport xmlns:ows="http://www.opengis.net/ows/1.1"
      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
      xsi:schemaLocation="http://www.opengis.net/ows/1.1 http://schemas.opengis.net/ows/1.1.0/owsExceptionReport.xsd"
      version="1.0.0" xml:lang="en">
      <ows:Exception exceptionCode="MissingParameterValue"
          locator="service">
        <ows:ExceptionText>The service parameter is missing</ows:ExceptionText>
      </ows:Exception>
    </ows:ExceptionReport>
    ```
  - old behaviour was either `internal error` or redirect to WMS Capabilities
- If the `service` parameter has an unsupported value (service is disabled or non existent), an Exception like this is thrown (with status code 400):
    ```
    <ows:ExceptionReport xmlns:ows="http://www.opengis.net/ows/1.1"
      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
      xsi:schemaLocation="http://www.opengis.net/ows/1.1 http://schemas.opengis.net/ows/1.1.0/owsExceptionReport.xsd"
      version="1.0.0" xml:lang="en">
      <ows:Exception exceptionCode="InvalidParameterValue"
          locator="service">
        <ows:ExceptionText>The value of the service parameter &quot;test&quot; is invalid</ows:ExceptionText>
      </ows:Exception>
    </ows:ExceptionReport>
    ```

XML-based Exceptions in general will now return with status code 500 as default instead of 200, which is the better choice in terms of conformity. This is a breaking change and needs a major version increase

### DIFF
--- a/mapproxy/service/templates/ows_exception.xml
+++ b/mapproxy/service/templates/ows_exception.xml
@@ -1,0 +1,10 @@
+<?xml version="1.0"?>
+<ows:ExceptionReport xmlns:ows="http://www.opengis.net/ows/1.1"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://www.opengis.net/ows/1.1 http://schemas.opengis.net/ows/1.1.0/owsExceptionReport.xsd"
+  version="1.0.0" xml:lang="en">
+  <ows:Exception{{if code is not None}} exceptionCode="{{code}}"{{endif}}
+     {{if locator is not None}} locator="{{locator}}"{{endif}}>
+    <ows:ExceptionText>{{exception}}</ows:ExceptionText>
+  </ows:Exception>
+</ows:ExceptionReport>
--- a/mapproxy/config/loader.py
+++ b/mapproxy/config/loader.py
@@ -24,6 +24,7 @@ from mapproxy.util.py import memoize
 from mapproxy.config.spec import validate_options, add_source_to_mapproxy_yaml_spec, add_service_to_mapproxy_yaml_spec
 from mapproxy.config.validator import validate
 from mapproxy.config import load_default_config, finish_base_config, defaults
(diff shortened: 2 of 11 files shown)


mapproxy/service/templates/ows_exception.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0"?>
+<ows:ExceptionReport xmlns:ows="http://www.opengis.net/ows/1.1"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://www.opengis.net/ows/1.1 http://schemas.opengis.net/ows/1.1.0/owsExceptionReport.xsd"
+  version="1.0.0" xml:lang="en">
+  <ows:Exception{{if code is not None}} exceptionCode="{{code}}"{{endif}}
+     {{if locator is not None}} locator="{{locator}}"{{endif}}>
+    <ows:ExceptionText>{{exception}}</ows:ExceptionText>
+  </ows:Exception>
+</ows:ExceptionReport>
```
hblitza, line 10:
```suggestion
</ows:ExceptionReport>

```

hblitza (approved):

mapproxy/service/templates/ows_exception.xml
weskamm: In this specific case, the non newline is intended :-)

weskamm:
Thanks for review!

PR #998, opened 2024-09-12
feat: handle file permissions for geopackage

hblitza:
Add handle file permissions of geopackage.

@weskamm


weskamm:
Could you give more details, whats the actual usecase?

weskamm:
Ah sorry, thought this was a feature request, now i see your changes. We will have to adapt all `sqlite.connect` calls to respect the permissions, there should be around 20 or 30 calls if i remember correctly.

PR #1006, opened 2024-10-18
feat: improve ensure_directory and add tests

simonseyock:
This PR simplifies the ensure_directory function and adds tests for it. Additionally it assumes that the permissions in the config are always given as a string to lessen the confusion about it being an base 8 or base 10 number.


hblitza (approved):

hblitza:
Nice work!

PR #1007, opened 2024-10-18
Permissions fix static legend source

simonseyock:
This PR fixes the static legend source config by using global permissions for the directory

### DIFF
--- a/mapproxy/config/loader.py
+++ b/mapproxy/config/loader.py
@@ -732,7 +732,13 @@ class WMSSourceConfiguration(SourceConfiguration):
             prefix = 'file://'
             url = prefix + context.globals.abspath(url[7:])
         lg_client = WMSLegendURLClient(url)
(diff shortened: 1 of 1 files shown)


mapproxy/config/loader.py
```diff
@@ -732,7 +732,13 @@ def static_legend_source(url, context):
             prefix = 'file://'
             url = prefix + context.globals.abspath(url[7:])
         lg_client = WMSLegendURLClient(url)
-        legend_cache = LegendCache(cache_dir=cache_dir, directory_permissions=CacheConfiguration.directory_permissions())
+
+        global_permissions=context.globals.get_value('directory_permissions', None,
+                                                          global_key='cache.directory_permissions')
```
hblitza, line 737: indentation correct here?

hblitza (approved):
LGTM, thanks!

mapproxy/config/loader.py
simonseyock: nah, it's incorrect. I'll fix it.

PR #1009, opened 2024-10-18
fix: always use cached connection, add test

simonseyock:
* always use cached connection in mbtiles cache
* add test


hblitza (approved):
Thanks for this!

PR #1012, opened 2024-10-18
doc: document permission configuration

simonseyock:

### DIFF
--- a/doc/caches.rst
+++ b/doc/caches.rst
@@ -90,7 +90,7 @@ This is the default cache type and it uses a single file for each tile. Availabl
   .. versionadded:: 2.0.0
 
 ``directory_permissions``, ``file_permissions``:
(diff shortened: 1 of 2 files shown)


doc/caches.rst
```diff
@@ -89,6 +89,11 @@ This is the default cache type and it uses a single file for each tile. Availabl
 
   .. versionadded:: 2.0.0
 
+``directory_permissions``, ``file_permissions``:
+  Permissions that MapProxy will set when creating files and directories. Must given as strings containing the octal representation of permissions. I.e. ``rwxrw-r--`` is ``'764'``.
```
weskamm, line 93:
```suggestion
  Permissions that MapProxy will set when creating files and directories. Must be given as string containing the octal representation of permissions. I.e. ``rwxrw-r--`` is ``'764'``. This will not work on windows OS.
```

simonseyock: good point

hblitza (approved):

PR #1013, opened 2024-10-18
Removed pin of werkzeug version in setup.py

### LINKED ISSUE #1000: Support newer versions of werkzeug?
mapproxy currently requires version 1.0.1 of werkzeug. This version is more than 4 years old, and has a number of security issues.

It'd be helpful if mapproxy supported newer versions of werkzeug.

Ideally it would be great to support version 3, but if that's not possible, than supporting 2.2.3 would be helpful.

ricardogsilva:
This PR lifts the pin on the werkzeug version `1.0.1`, as currently found on `setup.py`. It would allow MapProxy to be installed with recent versions of werkzeug, thus making it easier to use together with other components of the Python ecosystem.

---

- fixes #1000

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -10,7 +10,7 @@ install_requires = [
     'pyproj>=2',
     'jsonschema>=4',
     'importlib_resources;python_version<="3.8"',
-    'werkzeug==1.0.1'
+    'werkzeug'
 ]
 
 
--- a/setup.py
+++ b/setup.py
@@ -10,7 +10,7 @@ install_requires = [
     'pyproj>=2',
     'jsonschema>=4',
     'importlib_resources;python_version<="3.8"',
-    'werkzeug'
+    'werkzeug<4'
 ]
 
 


weskamm:
Thanks for your efforts, but Its not enough to remove the lock. Once greenkeeper starts an update, several tests and code will fail. Maybe you can have a look at that?

ricardogsilva:
@weskamm sure I can take a look

Which tests would fail though? These that ran on this PR all passed - If there are others that are not part of CI I can work on having those pass too, just let me know where I should look.

weskamm:
Well the tests on this PR did pass because no update happened. Just install a newer / latest version of werkzeug (`pip install werkzeug`) and run the tests (`pytest mapproxy/test`) or try to use mapproxy, it will fail. Its because werkzeug is used in some core functionailities. like the integrated dev server.

hblitza:
Allright, now I can also confirm that the tests run successfully using `werkzeug===3.0.4` locally. I also manually tested the serve-develop tool and it looked good. Concerning your change - I would keep a tested version lock, just like it is done for the other libs..

ricardogsilva:
@hblitza 

Respectfully I would disagree and would not put a locked version of any dependency in `setup.py`'s `install_requires` - the [Python packaging docs also mention it is not a good practice](https://packaging.python.org/en/latest/discussions/install-requires-vs-requirements/#install-requires) - they do recommend setting the upper/lower bounds of versions which are known to work, which I tend not to do, but must admit is a good idea :sweat_smile:.

In fact, in mapproxy's `setup.py`, `werkzeug` is the only dependency which was pinned, while all the others are following the best practice of simply indicating the range of acceptable versions.

With this in mind, I'm going to modify my PR to make the change be `werkzeug<4`, which I think would be safe - hopefully you will agree with this change.

hblitza (approved):

weskamm:
Thanks for your contributions and tests!

PR #1016, opened 2024-10-21
Fix transparency in TMS demo page

RichyHBM:
The logic for the demo page using TMS doesn't currently support transparency due to incorrect logic


hblitza (approved):

simonseyock (approved):

simonseyock:
Thanks for fix!

PR #1025, opened 2024-11-14
fix: use file permissions also for lock files

simonseyock:
The lock files were missing file permissions if set. This adds the functionality and adds a test for it.


hblitza (approved):
Thanks for this. LGTM (but untested)

PR #1042, opened 2024-11-22
fix: only use chmod if the file was created

simonseyock:
chmod can fail if the user is not the owner of the file. So we call it only if the file was created. I scanned for all calls of chmod.


marcjansen (approved):
Great thanks

PR #1047, opened 2024-11-26
feat: use python build instead of invoking setup.py

simonseyock:
Invoking setup.py to generate dist and sdist distributions is deprecated and generates errors on PyPI regarding PEP 625

> This email is notifying you of an upcoming deprecation that we have determined may affect you as a result of your recent upload to 'MapProxy'.
>
> In the future, PyPI will require all newly uploaded source distribution filenames to comply with PEP 625. Any source distributions already uploaded will remain in place as-is and do not need to be updated.
>
>Specifically, your recent upload of 'MapProxy-3.1.3.tar.gz' is incompatible with PEP 625 because it does not contain the normalized project name 'mapproxy'.
>
>In most cases, this can be resolved by upgrading the version of your build tooling to a later version that supports PEP 625 and produces compliant filenames.
>
> If you have questions, you can email admin@pypi.org to communicate with the PyPI admin@pypi.org to communicate with the PyPI administrators.

With the new system the issue should be resolved automatically

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -28,7 +28,8 @@ jobs:
 
       - name: Make wheel
         run: |
-          python setup.py egg_info -b "" -D sdist bdist_wheel
+          pip install -q build
+          pip -m build
 
       - name: Publish package to PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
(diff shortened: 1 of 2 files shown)


simonseyock:
I tested this locally and the sdist distribution is indeed then generated with a normalized name.

weskamm:
Thank you!

PR #1071, opened 2025-02-17
Update wmts_demo.html

tobias-brunner:
Fix WMTS-Demo Page (KVP seems not to be supported by mapproxy).

### DIFF
--- a/mapproxy/service/templates/demo/wmts_demo.html
+++ b/mapproxy/service/templates/demo/wmts_demo.html
@@ -46,12 +46,13 @@ jscript_functions=None
         const projection = ol.proj.get(srs);
 
         const source = new ol.source.WMTS({
-            url: '../service?',
+            url: '../wmts{{restful_url}}',
             layer: "{{layer.name}}",
             matrixSet: '{{matrix_set}}',
             format: "{{format}}",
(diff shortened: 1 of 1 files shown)


weskamm:
Did you read https://mapproxy.github.io/mapproxy/latest/services.html#kvp ?
What exactly did not work for you?

tobias-brunner:
Thanks for the quick reply! The demo page did not work for WMTS-services (using REST) after upgrading from 1.15 to 3.1.

No, I did not read that page. It is slightly ambiguous, as it says 

> You can enable or disable the KVP service with the kvp option. It is enabled by default and you need to enable restful if you disable this one.

but also 

> You can enable or disable the RESTful service with the restful option. It is enabled by default and you need to enable kvp if you disable this one.

As the use of KVP is hard-coded into the demo page, it really only works if KVP is enabled. Is that intentional? 

My fix obviously only relocates the problem from people using REST to those using KVP.
As far as my limited experience goes, REST is much more widely used than KVP, though.

A real fix would be to first check, if REST or KVP is enabled, and then use that protocol.

simonseyock:
I created a PR that adds this check: #1087 

Thanks for mentioning this issue and providing a solution for this!

PR #1087, opened 2025-03-21
Restful wmts demo

### LINKED ISSUE #1086: No ability for debug logging in production container.
I have tried several solutions but when i run the ”production ready” docker container i cant in any way enable debug logging.

Documentation writes about a log.ini which i have mounted but it does not get picked up.

Im making request through the demo app, but mapproxy logs nothing, the only logs i see are acces logs against the nginx.

How do you make mapproxy 3.1.3 to log its outgoing source requests? And other info so i can troubleshoot.

simonseyock:
Fixes the demo site for WMTS if kvp is disabled. Uses restful requests by default now.

Closes #1071


weskamm (approved):
Nice, thanks!

PR #1102, opened 2025-04-10
Security fixes

simonseyock:
Fixes:

- Security fix to prevent XSS injections in demo pages
- Security fix to prevent reading file urls

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,3 +1,12 @@
+4.0.2 2025-04-10
+~~~~~~~~~~~~~~~~
+
+Fixes:
+
+- Security fix to prevent XSS injections in demo pages
+- Security fix to prevent reading file urls
+
+
 4.0.1 2025-03-25
 ~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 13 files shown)


CHANGES.txt
```diff
@@ -1,3 +1,12 @@
+4.0.2 2025-04-10
+~~~~~~~~~~~~~~~~
+
+Fixes:
+
+- Security fix to prevent XSS injections in demmo pages
```
hblitza, line 6:
```suggestion
- Security fix to prevent XSS injections in demo pages
```

mapproxy/util/escape.py
```diff
@@ -0,0 +1,10 @@
+def escape_html(data):
```
weskamm, line 1: as it also escapes JS (at least it states), another method name could be used. but this is not blocking

weskamm (approved):

mapproxy/util/escape.py
simonseyock: I named it this way because it is used in html contexts ... More precise would be `escape_in_html`

weskamm: thanks for clarification

PR #1116, opened 2025-04-22
fix: use utf-8 in get feature info requests

simonseyock:
The GFI requests returned improperly html encoded strings. This should fix this.


hwbllmnn (approved):

simonseyock:
The GFI Requests to MapProxy sometimes returned bytes and sometimes strings. I changed this so MapProxy now always returns strings.

hwbllmnn (approved):

PR #1124, opened 2025-04-29
fix: round scale denominator to 10 digits after decimal point

simonseyock:
This rounds the scale denominator in the wmts capabilities to 10 digits after the decimal point.


weskamm (approved):
Works for me!

PR #1160, opened 2025-06-04
Optional tms metadata

hschiebold:
We would provide optional tms metadata KeywordList and ContactInformation for tms capabilities.
(https://wiki.osgeo.org/wiki/Tile_Map_Service_Specification)

We are doing it with code below since a few years for our customer.

i.e. https://via.bund.de/wsv/bwastr/tms/1.0.0


hschiebold:
Added simple test and full example in doc for second approach with standard keyword_list (we would prefer). Should we notice in doc, that if no special service tms md block is provided the wms md entries are used by default?

PR #1162, opened 2025-06-05
wms opts query_layers for featurinfo

hschiebold:
We would provide an wms_opts query_layers param for source of type wms. 
Under some circumstances we have to define a special Featurinfo layer of the source wms which we would cascade to wmts featureinfo. 

i.e. 

sources:
  BWASTR_wms:
    type: wms
    concurrent_requests: 4
    wms_opts:
       version: 1.1.1
       featureinfo: true
       query_layers: Gewaessernetz
       legendgraphic: true

https://via.bund.de/wsv/bwastr/wmts/1.0.0/WMTSCapabilities.xml

### DIFF
--- a/mapproxy/client/wms.py
+++ b/mapproxy/client/wms.py
@@ -180,7 +180,8 @@ class WMSInfoClient(object):
         req.params.pos = query.pos
         if query.feature_count:
             req.params['feature_count'] = query.feature_count
-        req.params['query_layers'] = req.params['layers']
+        if not 'query_layers' in req.params:
+            req.params['query_layers'] = req.params['layers']
         if 'info_format' not in req.params and query.info_format:
             req.params['info_format'] = query.info_format
         if not req.params.format:
--- a/mapproxy/config/loader.py
+++ b/mapproxy/config/loader.py
@@ -864,6 +864,8 @@ class WMSSourceConfiguration(SourceConfiguration):
             version = wms_opts.get('version', '1.1.1')
             if 'featureinfo_format' in wms_opts:
                 params['info_format'] = wms_opts['featureinfo_format']
+            if 'query_layers' in wms_opts:
+                params['query_layers'] = wms_opts['query_layers']
             fi_request = create_request(self.conf['req'], params,
                                         req_type='featureinfo', version=version,
                                         abspath=self.context.globals.abspath)


simonseyock:
This is a very nice addition, thanks!

Can you add a test and documentation for this feature?

The test can be very similar to this one: https://github.com/mapproxy/mapproxy/blob/5699e22d4cbcc1efe14220c09ea80ebf32cff764/mapproxy/test/system/test_wms.py#L634

The used mapproxy config lives here: https://github.com/mapproxy/mapproxy/blob/master/mapproxy/test/system/fixture/layer.yaml

Maybe you can just add another source and layer.

The documentation should go here I suppose: https://github.com/mapproxy/mapproxy/blob/5699e22d4cbcc1efe14220c09ea80ebf32cff764/doc/sources.rst?plain=1#L67

Also if you have time and know json-schema, please add the configuration to the validator: https://github.com/mapproxy/mapproxy/blob/5699e22d4cbcc1efe14220c09ea80ebf32cff764/mapproxy/config/config-schema.json#L674

The last is not mandatory as the validator is quite incomplete anyways.

mapproxy/test/system/test_wms.py
```diff
@@ -836,6 +837,24 @@ def test_get_featureinfo_not_queryable(self, app):
         assert "tms_cache is not queryable" in xml.xpath("//ServiceException/text()")[0]
         assert validate_with_dtd(xml, "wms/1.1.1/exception_1_1_1.dtd")
 
+    def test_get_featureinfo_query_layers(self, app):
+        expected_req = (
+            {
+                "path": r"/service?LAYERs=foo,bar&SERVICE=WMS&FORMAT=image%2Fpng"
+                "&REQUEST=GetFeatureInfo&HEIGHT=200&SRS=EPSG%3A900913"
+                "&VERSION=1.1.1&BBOX=1000.0,400.0,2000.0,1400.0&styles="
+                "&WIDTH=200&QUERY_LAYERs=foo&X=10&Y=20"
```
simonseyock, line 846:
```suggestion
                "&WIDTH=200&QUERY_LAYERS=foo&X=10&Y=20"
```

simonseyock (approved):
Thanks a lot. I will merge it soon.

PR #1177, opened 2025-06-17
feat: add build args to dockerfiles for user ids

simonseyock:


ahennr:
Changes included in #1426

PR #1186, opened 2025-06-23
restful encoding / style isdefault and urn:ogc:def:crs for SupportedCRS

hschiebold:
We provide a bit enhanced wmts capabilities doc with following template since a few years for our mapproxy wmts services to solve compability issues with some wmts clients.

We provide the restful encoding, the isDefault=true declaration for the one style and the urn:ogc:def:crs: prefix for EPSG:XXXX declarations of srs for compability.

We changed the DeliveryPoint to the more correct {{service.contact.address}} property (in our opinion).

Maybe it's usefull in general for others too - hopefully.

### DIFF
--- a/mapproxy/service/templates/wmts100capabilities.xml
+++ b/mapproxy/service/templates/wmts100capabilities.xml
@@ -43,7 +43,6 @@
     </ows:ServiceContact>
   </ows:ServiceProvider>
 {{endif}}
-{{if restful}}
   <ows:OperationsMetadata>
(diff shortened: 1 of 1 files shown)


simonseyock:
This looks like a good addition! Could you please change the template so it does not contain so much duplicate parts?

i.e. using this line:
```
<ows:Value>{{if restful}}RESTful{{else}}KVP{{endif}}</ows:Value>
```

instead of duplicating the whole `ows:OperationsMetadata` part.

PR #1198, opened 2025-07-08
Update mbtiles.py

SantiagoHorcajo:
This log.info line is missing the %s placeholder to replace the argument, and an error occurs when creating .mbtile files:

log.info("setting file permissions on MBTile file: ", permission)


weskamm:
Nice catch, thank you

PR #1266, opened 2025-09-15
Doc: fix warnings, and build doc as part of pull requests with warnings as errors

rouault:


weskamm (approved):
Seems the intention is to just make the pipeline fail when the documentaiton compilation fails. Good idea and looks good to me!

PR #1268, opened 2025-09-16
Add ogcapitiles and ogcapimaps source types

rouault:
Funded by Centre National d'Etudes Spatiales (CNES): https://cnes.fr

### DIFF
--- a/mapproxy/source/ogcapitiles.py
+++ b/mapproxy/source/ogcapitiles.py
@@ -159,7 +159,9 @@ class OGCAPITilesSource(MapLayer):
                 for tileset in map_tilesets_candidates:
                     tileset_links = tileset["links"]
                     for link in tileset_links:
-                        if "href" in link and self.tile_matrix_set_id in link["href"]:
+                        if "href" in link and link["href"].split("?")[0].endswith(
+                            "/" + self.tile_matrix_set_id
+                        ):
                             user_matrix_sets.append(tileset)
                             break
                 map_tilesets_candidates = user_matrix_sets


mapproxy/source/ogcapimaps.py
```diff
@@ -0,0 +1,189 @@
...
+
+            if "links" not in j:
+                ex = SourceError(f"Could not retrieve 'links' in {url} response")
+                log.error(ex)
+                raise ex
+
+            for mimetype in ("image/png", "image/jpeg"):
+                href = find_href_in_links(
+                    j["links"],
+                    "http://www.opengis.net/def/rel/ogc/1.0/map",
```
simonseyock, line 121: According to the standard, the value can also be `[ogc-rel:map]`: https://docs.ogc.org/is/20-058/20-058.html#req_collection-map_desc-links

simonseyock: If no link with rel=http://www.opengis.net/def/rel/ogc/1.0/map or rel=[ogc-rel:map] is found, MapProxy should raise a NOTICE or WARNING that the specified collection does not contain an OGC API Maps endpoint.

doc/sources.rst
```diff
@@ -646,3 +648,118 @@ Example:
   debug_source:
     type: debug
 
+
+.. _ogcapitiles_label:
+
+OGC API Tiles
+"""""""""""""
+
+.. versionadded:: 5.1.0
```
weskamm, line 657: As we do have version specific documentations, you can omit the `versionadded`, this is not needed anymore.

doc/sources.rst
```diff
@@ -646,3 +648,118 @@ Example:
...
+
+Use the type ``ogcapitiles`` to for servers implementing `OGC API Tiles <https://docs.ogc.org/is/20-057/20-057.html>`__.
```
weskamm, line 659:
```suggestion
Use the type ``ogcapitiles`` for servers implementing `OGC API Tiles <https://docs.ogc.org/is/20-057/20-057.html>`__.
```

doc/sources.rst
```diff
@@ -646,3 +648,118 @@ Example:
...
+    landingpage_url: http://example.com/ogcapi
+    collection: my_collection_name
+
+
+.. _ogcapimaps_label:
+
+OGC API Maps
+""""""""""""
+
+.. versionadded:: 5.1.0
```
weskamm, line 705: see above, please remove

doc/sources.rst
```diff
@@ -646,3 +648,118 @@ Example:
...
+
+Use the type ``ogcapimaps`` to for servers implementing `OGC API Maps <https://docs.ogc.org/is/20-058/20-058.html>`__.
```
weskamm, line 707:
```suggestion
Use the type ``ogcapimaps`` for servers implementing `OGC API Maps <https://docs.ogc.org/is/20-058/20-058.html>`__.
```

doc/sources.rst
```diff
@@ -646,3 +648,118 @@ Example:
...
+
+[optional]  Can be set to ``true`` if you want to use this source as an overlay.
+
+``image``
+^^^^^^^^^
+
+[optional] 
+
+See :ref:`image_options` for other options.
+
```
weskamm, line 736:
```suggestion

Additionally to the image options, the following parameters can be used:

```

mapproxy/source/ogcapitiles.py
```diff
@@ -0,0 +1,367 @@
...
+            configuration_context = copy.copy(self.configuration_context)
+            configuration_context.grids = {grid.name: WrappedGridConfig(grid)}
+            configuration_context.sources = {"this_source": WrappedSourceConfig(self)}
+            cache_config = CacheConfiguration(cache_conf, configuration_context)
+            _, _, tile_manager = cache_config.caches()[0]
+            cacheMapLayer = CacheMapLayer(tile_manager, image_opts=self.image_opts)
+
+            self.map_grid_name_to_cachemaplayer[grid.name] = cacheMapLayer
+
+            return cacheMapLayer
```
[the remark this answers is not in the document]
weskamm: We would favor a solution that places your (useful!) code into mapproxy-util. This is a better place for configuration helpers and keeps mapproxy behave the same in all places - manually configurable

mapproxy/grid/tile_grid.py
```diff
@@ -83,6 +83,87 @@ def tile_grid(srs=None, bbox=None, bbox_srs=None, tile_size=(256, 256),
...
+                        y - tileHeight * matrixHeight * cellSize,
+                        x + tileWidth * matrixWidth * cellSize,
+                        y]
+            elif cornerOfOrigin == "bottomLeft":
+                origin = 'll'
+                bbox = [x, y,
+                        x + tileWidth * matrixWidth * cellSize,
+                        y + tileHeight * matrixHeight * cellSize]
+            else:
+                raise UnsupportedException(f"Invalid cornerOfOrigin={cornerOfOrigin}")
```
weskamm, line 138: It would be nice if this could be refactored to its own method for better testing

doc/sources.rst
```diff
@@ -646,3 +648,118 @@ Example:
   debug_source:
     type: debug
 
+
+.. _ogcapitiles_label:
+
+OGC API Tiles
```
simonseyock, line 654: This section should contain a prominently placed notice/warning that only map tiles and no vector tiles are supported.

mapproxy/source/ogcapimaps.py
```diff
@@ -0,0 +1,189 @@
...
+                query.bbox[1],
+                query.bbox[2],
+                query.bbox[3],
+            )
+        if query.srs.srs_code != "EPSG:4326":
+            req.params["bbox-crs"] = self.map_crs_code_to_url[query.srs.srs_code]
+        if not (
+            query.srs.srs_code == "EPSG:4326"
+            and len(self.supported_srs.supported_srs) == 1
+        ):
```
weskamm, line 175:
```suggestion
        if query.srs.srs_code != "EPSG:4326" or len(self.supported_srs.supported_srs) != 1:
```

mapproxy/source/ogcapimaps.py
```diff
@@ -0,0 +1,189 @@
...
+        req.params["height"] = query.size[1]
+        if self.image_opts.transparent:
+            req.params["transparent"] = "true"
+        if self.transparent_color:
+            if len(self.transparent_color) == 4:
+                r, g, b, a = self.transparent_color
+                req.params["bgcolor"] = "0x%02X%02X%02X%02X" % (a, r, g, b)
+            else:
+                r, g, b = self.transparent_color
+                req.params["bgcolor"] = "0x%02X%02X%02X" % (r, g, b)
```
weskamm, line 187: transparent_color is used in mapproxy to specify a color that should be output by mapproxy as transparent (by internal image manipulation). See `WMSLikeSource`, mehtod `make_transparent`.

The bgcolor on the opposite is used to specify a color for transparent / no data values in the source, that should be painted in the given color.

The code can be removed, as the handling of transparent_color should be handled already in the parent class.
Or this could be rewritten to just support and passthrough the parameter `bgcolor` from the user config to the source as is.

mapproxy/source/ogcapimaps.py
```diff
@@ -0,0 +1,189 @@
...
+    build_absolute_url,
+)
+from mapproxy.util.py import reraise_exception
+
+import logging
+
+log = logging.getLogger("mapproxy.source.ogcapimaps")
+
+# For testing
+reset_cache = False
```
simonseyock, line 39: I am not entirely happy with this global variable, but I can't think of a better way right now. But if you use it, please name it differently as the word `cache` is quite loaded in the MapProxy context.

mapproxy/source/ogcapitiles.py
```diff
@@ -0,0 +1,367 @@
...
+                reraise_exception(SourceError(e.args[0]), sys.exc_info())
+
+            if "links" not in j:
+                ex = SourceError(f"Could not retrieve 'links' in {url} response")
+                log.error(ex)
+                raise ex
+
+            tilesets_map_href = find_href_in_links(
+                j["links"],
+                "http://www.opengis.net/def/rel/ogc/1.0/tilesets-map",
```
weskamm, line 132: according to the spec, the value `[ogc-rel:tilesets-map]` should also be allowed / supported

mapproxy/source/ogcapitiles.py
```diff
@@ -0,0 +1,367 @@
...
+                    f"Could not retrieve 'tilesets' in {tilesets_map_url} response"
+                )
+                log.error(ex)
+                raise ex
+
+            for tileset in j["tilesets"]:
+                if tileset["dataType"] != "map":
+                    continue
+
+                ok = False
```
weskamm, line 167:
```suggestion
                user_matrix_set_found = False
```

mapproxy/source/ogcapitiles.py
```diff
@@ -0,0 +1,367 @@
...
+                if self.tile_matrix_set_id:
+                    tileset_links = tileset["links"]
+                    for link in tileset_links:
+                        if "href" in link and self.tile_matrix_set_id in link["href"]:
+                            ok = True
+                            break
+                else:
+                    ok = True
```
weskamm, line 175:
```suggestion
```

mapproxy/source/ogcapitiles.py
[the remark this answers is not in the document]
simonseyock: Another possibility would be to add the option to configure a grid with just an ogc landing page url and optional tileset_matrix_id.

But all these autoconfiguring via http request have one major problem: As soon as the source becomes unavailable the cache becomes unusable.

Another problem occurs if the user wants to restrict the grid. i.e. only wants to offer certain resolutions.

The easiest way to persist the grid config is via config and the mapproxy-util. That is the reason we favor this solution.

weskamm (changes requested):
Thanks for the contribution! This looks really promising and our tests have been successfull so far. It seems to be a very good base for the intetgration of OGC API sources! 

We left some more or less minor comments. The automatic grid and cache configuration that happens in the `ogcapitiles` class should be changed or further discussed.

PR #1284, opened 2025-09-27
Add support for OGC API Maps and Tiles service (server side)

rouault:
(on top of PR #1268)

Funded by Centre National d'Etudes Spatiales (CNES): https://cnes.fr/


mapproxy/service/ogcapi/map.py
```diff
@@ -0,0 +1,845 @@
...
+            "transparent",
+        ]
+    )
+    collections = req.args.get("collections", None)
+
+    if coll_id is None and (server.default_dataset_layers or collections):
+        allowed_args.add("collections")
+
+    if req.args.get("service", None) == "WMS":
+        # Hack for the Leaflet map in our collection.html that issues WMS requests...
```
weskamm, line 800: :-) We would prefer to use openlayers here, as its already used in the other demo applications and makes it easier to maintain when only having one clientside mapping library.
Support for ogc api map tiles is already there, ogc api maps can be added via a custom loader function. Ideally, the preview map requests the layer the way the source is configured in mapproxy, e.g. if its configured as ogcapitiles, it would be correctly requested like that.

mapproxy/service/ogcapi/map.py
```diff
@@ -0,0 +1,845 @@
...
+            )
+
+        if bbox[idx_min] is not None:
+            raise OGCAPIServer.invalid_parameter(
+                f"Axis name {axis_name} has been specified in multiple subset parts"
+            )
+
+        try:
+            bbox[idx_min] = float(range_part[0])
+            bbox[idx_max] = float(range_part[1])
```
simonseyock:
```suggestion
            bbox[idx_min] = float(low)
            bbox[idx_max] = float(high)
```

mapproxy/service/ogcapi/map.py
```diff
@@ -0,0 +1,619 @@
...
+    if query != orig_query:
+        result = SubImageSource(
+            result, size=orig_query.size, offset=offset, image_opts=image_opts
+        )
+
+    # Provide the wrapping WSGI app or filter the opportunity to process the
+    # image before it's wrapped up in a response
+    result = server.decorate_img(
+        result,
+        "wms.map",
```
weskamm:
```suggestion
        "ogcapi.map",
```

mapproxy/service/ogcapi/map.py
```diff
@@ -0,0 +1,619 @@
...
+            "transparent",
+        ]
+    )
+    collections = req.args.get("collections", None)
+
+    if coll_id is None and (server.default_dataset_layers or collections):
+        allowed_args.add("collections")
+
+    if req.args.get("service", None) == "WMS":
+        # Hack for the OpenLayers map in our collection.html that issues WMS requests...
```
weskamm: We will need a fixup for this in a follow up PR, just adding comment as a reminder for myself

mapproxy/service/templates/ogcapi/openapi/redoc.html
```diff
@@ -0,0 +1,19 @@
...
+    <style>
+      body {
+        margin: 0;
+        padding: 0;
+      }
+    </style>
+  </head>
+  <body>
+    <redoc spec-url='{{ data['openapi-document-path'] }}?f=json'></redoc>
+    <script src="https://cdn.jsdelivr.net/npm/redoc@next/bundles/redoc.standalone.js"> </script>
```
weskamm: please also serve this from the static folder

mapproxy/service/templates/ogcapi/openapi/swagger.html
```diff
@@ -0,0 +1,57 @@
+<!-- HTML for static distribution bundle build -->
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8">
+    <title>Swagger UI - {{ config['metadata']['identification']['title'] }}</title>
+    <link rel="stylesheet" type="text/css" href="https://unpkg.com/swagger-ui-dist/swagger-ui.css" >
+    <link rel="icon" type="image/png" href="https://unpkg.com/swagger-ui-dist/favicon-32x32.png" sizes="32x32" />
+    <link rel="icon" type="image/png" href="https://unpkg.com/swagger-ui-dist/favicon-16x16.png" sizes="16x16" />
```
weskamm: please also serve this from the static folder

mapproxy/config/loader.py
```diff
@@ -2401,6 +2414,63 @@ def wms_service(self, conf):
...
+        image_formats = OrderedDict()
+        for format in image_formats_names:
+            opts = self.context.globals.image_options.image_opts({}, format)
+            if opts.format in image_formats:
+                log.warning('duplicate mime-type for WMS image_formats: "%s" already configured, will use last format',
+                            opts.format)
+            image_formats[opts.format] = opts
+
+        max_output_pixels = self.context.globals.get_value('max_output_pixels', conf,
+                                                           global_key='wms.max_output_pixels')
```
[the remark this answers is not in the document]
rouault: fixed in 4523527bd7dfdd199c0c4794fbc1ceaf0024235d

rouault:
> Thank you so far, this is nearly good to go. There is only the remaining issue with the `compatible_srs` left. Can you tackle that?

I'm repasting here a question I left yesterday below one of your comments and that probably got lost in the middle of the notification traffic:    
"We could potentially remove the compatible_srs property at the layer level if all sources would have a supported_srs property. But that is not the case currently. Some sources have a supported_srs member, some don't. A way of solving this would be that the MapLayer base class should have a self.supported_srs member (not necessarily populated by all sources, but still there and initialized to an empty list by default). Should I do that?"

simonseyock:
> > Thank you so far, this is nearly good to go. There is only the remaining issue with the `compatible_srs` left. Can you tackle that?
> 
> I'm repasting here a question I left yesterday below one of your comments and that probably got lost in the middle of the notification traffic: "We could potentially remove the compatible_srs property at the layer level if all sources would have a supported_srs property. But that is not the case currently. Some sources have a supported_srs member, some don't. A way of solving this would be that the MapLayer base class should have a self.supported_srs member (not necessarily populated by all sources, but still there and initialized to an empty list by default). Should I do that?"

Hey, I commented about this issue, maybe you did not see this:

At the moment MapProxy does not allow to restrict SRS on layer level. Instead it is configured on the service level. We would prefer to keep it there for now and not allow a different configuration on layer level.

See https://mapproxy.github.io/mapproxy/latest/services.html#srs

for example:

```yaml
services:
  wms:
    srs: ['EPSG:4326', 'EPSG:1234']
  ogcapi:
    map_srs: ['EPSG:3857', 'EPSG: 5678']
```

The additional option on layer level offers a granularity that is not needed at our opinion (at least there never was a feature request in that regard) and only adds more complexity.

simonseyock:
For example in the wms service configuration, it is read out as follows:
```python
        srs = self.context.globals.get_value('srs', conf, global_key='wms.srs')
        self.context.globals.base_config.wms.srs = srs
```

rouault:
@simonseyock  To elaborate further on why I was trying to have a per-layer list of compatible CRS is that one of reasonable use case is for a single mapproxy installation to be able to deliver data for planetary datasets of different planets. Having a per-service list of allowed CRS is a bit annoying since users could potentially try to select a CRS valid for Mars with Moon datasets. Of course I agree not having this per-layer granularity is not totally blocking (if they try to do something stupid they'll get an error), just that what I proposed to add seemed to me to be more user friendly in avoiding them to do non-sensical request. But I understand your concern of added complexity and confusion for administrators by adding a new configuration item that looks similar to an existing one

weskamm:
Thanks for your contributions and patience in the discussion! This looks really good and is a big step forward for mapproxy to be future compatible. This will be released soon with the next major version!

weskamm:
> > This will be released soon with the next major version!
> 
> Excellent. Do you have any timeframe for that? I might come with follow-up PRs depending on the feedback of the end user

Not yet, but the next week may be a realistic time

PR #1292, opened 2025-10-01
fix(ows): make service non mandatory for WMS GetMap requests

simonseyock:
According to spec the service parameter is not mandatory for WMS GetMap requests.

### DIFF
--- a/mapproxy/service/ows.py
+++ b/mapproxy/service/ows.py
@@ -36,9 +36,11 @@ class OWSServer(object):
     def handle(self, req):
         service = req.args.get('service')
         wmtver = req.args.get('wmtver')
+        request_arg = req.args.get('request')
         if not service:
-            if wmtver == '1.0.0':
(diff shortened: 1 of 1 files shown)


mapproxy/service/ows.py
```diff
@@ -36,9 +36,11 @@ def __init__(self, services):
     def handle(self, req):
         service = req.args.get('service')
         wmtver = req.args.get('wmtver')
+        request_arg = req.args.get('request')
         if not service:
-            if wmtver == '1.0.0':
+            if wmtver == '1.0.0' or request_arg == 'GetMap':
                 # WMS version 1.0.0 did not have a mandatory service parameter
+                # and for the `GetMap` request the service also is not mandatory
```
weskamm, line 43: so is the `wmtver` parameter optional? beacuse if not, this seems unnecessary

weskamm (approved):

PR #1294, opened 2025-10-03
Use correct command for Waitress serve

symroe:
The actual command is `waitress-serve` not `waitress`


simonseyock:
Thanks for your contribution!

Issue #556, opened 2021-12-16
transparent_color doesn't apply when mixing different sorces in a source


trucheromayor:
¡Congratulations ana Thanks for your great sofwware!

I hope you can attend to the following possible issue.

## Summary

When you define a transparent color in a simple source it works nicely.  But when you define a source composed of different sources it doesn't.

## Context
A mapproxy configuration like the following.
```yaml
layers:
  - name: OI_Copernicus_core003
    title: Copernicus Core003 Seamline
    sources: [core_cache]

  - name: Layer_mixture
    title: Layer_mixture
    sources: [core_cache,OI_cache_AT_2020]

  - name: Cache_mixture
    title: Cache_mixture
    sources: [Cache_mixture]

caches:
  Cache_mixture:
    grids:
    - EPSG3857
    image:
      resampling_method: nearest
    sources: [core_cache,OI_cache_AT_2020]
  core_cache:
    grids:
    - EPSG3857
    sources: [core003_wms]
  OI_cache_AT_2020:
    grids:
    - EPSG3857
    image:
      resampling_method: nearest
    sources: [OI_AT_2020_wms]

sources:
  core003_wms:
    type: wms
    min_scale: 50000
    max_scale: 1000000000
    wms_opts:
         version: 1.3.0
    image:
      transparent_color: '#ffffff'
      transparent_color_tolerance: 20
    req:
      url: https://jeodpp.jrc.ec.europa.eu/jeodpp/services/ows/wms/copernicus/core003?
      layers: OI.Mosaic.NaturalColor.Seamline
      transparent: true

  OI_AT_2020_wms:
    type: wms
    max_scale: 50000
    min_scale: 500
    concurrent_requests: 6
    wms_opts:
         version: 1.3.0
    image:
      transparent_color: '#ffffff'
      transparent_color_tolerance: 20
    req:
      url: https://wsa.bev.gv.at/GeoServer/Interceptor/Wms/OI/INSPIRE_KUNDEN-ea22c974-a9cc-4ef4-93b6-911aca065d6c?SERVICE=WMS&TIME=2020-01-01/2020-12-31
      layers: AT_BEV_OI
      transparent: true
    on_error:
      other:
        response: transparent
        cache: False
```
## Expected Behavior
The mixed cache doesn't apply transparent color

## Actual Behavior
This is the resulting demo
[...]

PR #1295, opened 2025-10-03
Prevent white pixels instead of transparency when merging sources

damien-faivre:
Fixed by @tobwen in this issue 👉 https://github.com/mapproxy/mapproxy/issues/556

### DIFF
--- a/mapproxy/cache/tile.py
+++ b/mapproxy/cache/tile.py
@@ -456,7 +456,7 @@ class TileCreator(object):
             except BlankImage:
                 return None
 
-        def get_map_from_source(source):
+         def get_map_from_source(source):
             try:
                 img = source.get_map(query)
             except BlankImage:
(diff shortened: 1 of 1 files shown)


mapproxy/cache/tile.py
```diff
@@ -465,9 +465,17 @@ def get_map_from_source(source):
...
         for layer in async_.imap(get_map_from_source, self.sources):
             if layer[0] is not None:
                 layers.append(layer)
+                try:
+                    transparent_layers.append(layer[0].image_opts.transparent)
+                except AttributeError:
+                    transparent_layers.append(None)
+
+        if len(transparent_layers) > 0 and all(transparent_layers):
+            self.tile_mgr.image_opts.transparent = True
```
simonseyock:
```suggestion
        all_transparent = True
        for layer in async_.imap(get_map_from_source, self.sources):
            if layer[0] is not None:
                layers.append(layer)
                if 'image_opts' in layer[0] and layer[0].image_opts.transparent:
                     all_transparent = all_transparent and layer[0].image_opts.transparent
                else: 
                     all_transparent = False

        if all_transparent:
            self.tile_mgr.image_opts.transparent = True
```

please check if this works. The code is a little clearer than using `except`.

simonseyock (changes requested):
Thanks for the contribution!

mapproxy/cache/tile.py
damien-faivre: Thanks for your feedbacks, and thanks again @tobwen for this code snippet. It works perfectly fine as is but we plan to test your suggestion Simon. Will let you know in a few days.

tobwen: I think this one was it: https://github.com/tobwen/mapproxy/commit/cbc3ccaee92996fccb23a51873038e4c462aca23

I hope I can test it tomorrow.

PR #1339, opened 2025-10-30
register_service_configuration(): add an optional schema_service parameter

rouault:
Since the introduction of this method, JSON Schema validation was added. When a plugin adds a new service which requires configuration elements in the .yaml configuration file, it is also needed to extend the JSON schema to avoid warnings at startup. Will be used by https://github.com/rouault/mapproxy_hips.


simonseyock:
Hello @rouault, thanks for your contribution! We merged a small refactoring where we splitted the `loader.py` file into multiple files. It would be nice if you could rebase your PR.

PR #1348, opened 2025-11-20
Fix OGC API: use req.script_url so SCRIPT_NAME is respected behind re…

### LINKED ISSUE #1344: OGC API links ignore SCRIPT_NAME behind reverse proxy (unlike WMS)
### Summary

When MapProxy is deployed behind a reverse proxy under a subpath (e.g., `/mapproxy`), the OGC API endpoints generate incorrect absolute URLs, because they do not respect `SCRIPT_NAME`.

Example request:

    https://myserver.com/mapproxy/ogcapi/collections?f=json

OGC API returns links like:

    http://myserver.com/ogcapi/collections/<collection-id>?f=json

The prefix `/mapproxy` is missing.

However, WMS GetCapabilities generates correct URLs, including the prefix:

    <OnlineResource xlink:href="https://myserver.com/mapproxy/service"/>

This results in inconsistent behavior across services.

---

### Cause

The issue originates from:

    OGCAPIServer.create_href()

which currently builds URLs using:

    req.host_url

`host_url` does not honor `SCRIPT_NAME`, so the external prefix is lost.

---

### Expected Behavior

All OGC API links should include the full external prefix, just like WMS already does:

    https://myserver.com/mapproxy/ogcapi/...

---

### Proposed Solution

Modify `OGCAPIServer.create_href()` to use:

    req.script_url

instead of:

    req.host_url

`script_url` already contains:

- scheme
- host
- SCRIPT_NAME prefix

Thus it generates correct proxied URLs.

Minimal fix example:

    def create_href(self, req, resource):
        return req.script_url + resource

---

### Why This Matters

- OGC API responses include absolute URLs required by modern web clients (OpenLayers, Cesium, etc.).
- Incorrect links break deployments behind reverse proxies.
- WMS/WMTS already handle `SCRIPT_NAME` correctly; OGC API should be consistent.

---

sbernads:
### Fix: OGC API does not respect SCRIPT_NAME behind reverse proxies

This PR replaces `req.host_url` with `req.script_url` inside  
`OGCAPIServer.create_href()`.

`script_url` already includes:

- scheme  
- host  
- external prefix (`SCRIPT_NAME`)  

so OGC API links become consistent with WMS/WMTS URLs, which already respect `SCRIPT_NAME`.

This fixes deployments where MapProxy runs behind a reverse proxy under a path prefix (for example `/geo`), where OGC API currently produces incorrect URLs such as:

```text
https://example.com/ogcapi/collections/<id>

instead of:

https://example.com/<prefix>/ogcapi/collections/<id>
```
Minimal patch applied:
```python
def create_href(self, req, resource):
    return req.script_url + resource
```

Fixes #1344

### DIFF
--- a/mapproxy/service/ogcapi/server.py
+++ b/mapproxy/service/ogcapi/server.py
@@ -35,6 +35,8 @@ from mapproxy.service.ogcapi.constants import (
 )
 from mapproxy.util.jinja2_templates import render_j2_template
 
+from mapproxy.util.escape import escape_html
+
 import logging
 
 log = logging.getLogger(__name__)


mapproxy/service/ogcapi/server.py
```diff
@@ -282,7 +282,7 @@ def handle(self, req: Request):
             )
 
     def create_href(self, req: Request, resource):
-        return req.host_url[:-1] + resource
+        return req.script_url + resource
```
simonseyock:
```suggestion
        return escape_html(req.script_url) + resource
```

PR #1362, opened 2026-01-11
feat: remove python 3.9 support

simonseyock:
Python 3.9 reached end-of-life


simonseyock:
I removed the python 3.14 support as this will need rework of many classes that are passed to subthreads.

Issue #1317, opened 2025-10-20
Use OGCMapsAPI for map output in the html of the OGC API


simonseyock:
At the moment the ogcapi html pages use the WMS service to display their maps. If a WMS is not available this causes problems.

It would be ideal if we could use a dedicated class from OpenLayers: https://github.com/openlayers/openlayers/issues/17150

PR #1417, opened 2026-04-10
Feature: Update OpenLayers and use OGCMapTile source in OGC-API demo pages

ahennr:
Since OpenLayers contains a source for OGCMaps API now (see https://github.com/openlayers/openlayers/issues/17150), this can be used in OGC Maps HTML preview pages.

In this context, a bug in location of `proj4` definitions has been fixed as well.

This PR adds this funtionality by upating OpenLayers to version 10.9.0 and replacing the corresponding OL sources.

See also https://github.com/mapproxy/mapproxy/issues/1317.


hwbllmnn (approved):

mapproxy/service/ogcapi/collections.py
```diff
@@ -62,10 +62,11 @@ def get_collection(server: OGCAPIServer, req: Request, layer: WMSLayerBase):
                         "attribution"
                     ] = f"[![{attribution_title}]({attribution_logo_url}]({attribution_url})"
                 else:
-                    col["attribution"] = f"[{attribution_title}]({attribution_url})"
+                    col[
+                        "attribution"] = f"[{attribution_title}]({attribution_url})"
```
marcjansen: Why do we introduce a newline here?

marcjansen (commented):
Only one question, non-blocking

mapproxy/service/ogcapi/collections.py
ahennr: A newline was added accidentally and is not intentional, sry. Removed in dcf5263

marcjansen (approved):
Great, thanks.

PR #1427, opened 2026-05-05
feat(docker): tunable UWSGI_PROCESSES and UWSGI_THREADS, vacuum=true

dietrichmax:
The nginx image hardcodes `processes` and `threads` in `docker/uwsgi.conf`, so tuning them today requires forking the image. This exposes both as env vars using uWSGI's native `$(VAR)` ini expansion, with defaults exported in `run-nginx.sh`:

- `UWSGI_PROCESSES` - default `2`
- `UWSGI_THREADS` - default `10`

Existing deployments are unchanged (defaults match the previous hardcoded values). Also enables `vacuum = true` so uWSGI removes its socket and pidfile on shutdown.

Verified locally:

```bash
docker build -f Dockerfile --target nginx -t mapproxy-nginx-pr .

# defaults — spawns 2 workers, cores: 10
docker run --rm -p 80:80 mapproxy-nginx-pr

# overrides — spawns 4 workers, cores: 20
docker run --rm -p 80:80 \
  -e UWSGI_PROCESSES=4 -e UWSGI_THREADS=20 \
  mapproxy-nginx-pr
```


simonseyock:
Can't you just mount a uwsgi.conf to `/mapproxy/uwsgi.conf` with adjusted values?

dietrichmax:
Yes, that's also possible and this PR doesn't break that path. Users who want to tune `max-requests`, `harakiri`, etc. can still mount a full config.

I went with env vars because the most common case is just "I want 4 workers instead of 2" and `-e UWSGI_PROCESSES=4` is way less friction than maintaining a full config copy. The mount approach requires keeping a full copy of uwsgi.conf in the deployment and re-syncing it whenever upstream changes, e.g. anyone using a mounted override would lose the `vacuum = true` after the merge if not updated manually.

But it also "opens the door" for several other configuration env vars that could be potentially added which I think is fine for a software like mapproxy. E.g. the [docker image of geoserver](https://github.com/geoserver/docker#environment-variables) also offers several configuration options via env vars.

That said, if you'd rather skip env vars entirely and just document `/mapproxy/uwsgi.conf` as the supported override point, I'm happy to close this and open a docs-only PR instead.

ahennr (approved):
Personally, I find the changes useful and support them. After all, they expand on the current configurations. If you want to completely override the `uwsgi` config, you can do that as you said @simonseyock

simonseyock:
I don't think the `vacuum` is needed. It only applies if the service gets stopped. But this should always lead to a stop of the docker image and therefore closing all sockets anyways.

simonseyock:
If you can convince me that `vacuum` is useful, I will add it swiftly.

simonseyock:
Thank you very much for your contribution.

PR #1431, opened 2026-05-19
Fix: Remove with-statement in seeder as ProxyConfiguration is not a contextmanager

NicoManderySWM:
In bd8f011ec302f3be1f3dc3a63f967b3db8c41bad the contextmanager methods where removed from the ProxyConfiguration class. The seeding script still expected this protocol.


hwbllmnn:
We think this should be fixed with #1442 . We'll release 6.1.1 shortly, can you test if that fixes your issue?

NicoManderySWM:
The solution from #1442 to keep the context manager interface works for me as well. Closing this issue. Thank you.

Issue #1316, opened 2025-10-20
FeatureInfo parsing for bad HTML


simonseyock:
We have updated lxml in MapProxy. lxml in the newer versions is much more strict. This might cause issues with feature info services all around. Missing closing tags or multiple starting html tags are common mistakes in many services. Browsers are very lenient in their interpretation of HTML, maybe their exists a way to leverage this functionality.

ahennr:
Fixed in #1432

PR #1432, opened 2026-05-27
Improve robustness of FeatureInfoDoc parsing and serialization for HTML / XML

ahennr:
In this PR, the robustness of FeatureInfoDoc parsing and serialization for HTML / XML is improved. Instead of raising exceptions on bad input, the code now logs warnings and returns safe empty defaults, allowing combine operations to skip invalid documents gracefully.

**Changes**

* Consolidated content validation and unified ElementTree return types across `XMLFeatureInfoDoc` and `HTMLFeatureInfoDoc`
* Fixed double UTF-8 encoding of special characters during HTML serialization
* Extended test coverage for edge cases: empty input, bytes input, broken HTML, and empty documents within combine() calls
* Upgraded lxml `6.0.2` → `6.1.1`

Please review

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,6 @@ nosetests*.xml
 .cursor/
 .claude/
 .vscode/
+
+# Virtual environments
+venv/
--- a/mapproxy/featureinfo.py
+++ b/mapproxy/featureinfo.py
@@ -15,13 +15,16 @@
 
 from codecs import decode
 import copy
+import logging
 import json
 
 from functools import reduce
-from io import StringIO, BytesIO
+from io import BytesIO
 from typing import Optional, Union
 
 from lxml import etree, html
+
+log = logging.getLogger('mapproxy.featureinfo')
 
 
 class FeatureInfoDoc(object):
@@ -49,7 +52,8 @@ class TextFeatureInfoDoc(FeatureInfoDoc):
 
     @classmethod
     def combine(cls, docs: list[FeatureInfoDoc]):
-        result_content = [doc.as_string() for doc in docs]
+        result_content = [s for doc in docs if (s := doc.as_string())]
+
         return cls("\n".join(result_content))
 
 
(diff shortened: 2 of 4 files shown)


ahennr:
Fixes issue #1316

marcjansen (approved):
Very nice work @ahennr :clap: , this looks good to me. Test coverage of the touched source file is great with 91 %; so I am quite confident this is fine.

This is a behavioural change, even though it fixes a bug. Can we think of some instance, where the new behaviour introduces a regression for users? I cannot come up with an idea, but maybe we should have this in the upgrade notes, so that users know what to look for e.g. in log files.

Just my 2 ct.

hwbllmnn (approved):

ahennr:
> This is a behavioural change, even though it fixes a bug. Can we think of some instance, where the new behaviour introduces a regression for users? I cannot come up with an idea, but maybe we should have this in the upgrade notes, so that users know what to look for e.g. in log files.

That's right; we should include the following in the release notes for the upcoming release @marcjansen ?

XML / HMTL errors in GFI requests are handled in a more lenient way now. Instead of  raising exceptions on bad input, the code now logs warnings and returns safe empty defaults. In log messages, you'll find hint like these:

```
[2026-05-28 12:08:12,474] mapproxy.featureinfo - WARNING - Failed to parse content as XML / HTML, returning empty element tree
[2026-05-28 12:08:12,474] mapproxy.featureinfo - WARNING - Document has no body element, skipping: 
```

marcjansen:
> That's right; we should include the following in the release notes for the upcoming release @marcjansen ?
> 

Great, thanks. I like it.


> XML / HMTL errors in GFI requests are handled in a more lenient way now. Instead of raising exceptions on bad input, the code now logs warnings and returns safe empty defaults. In log messages, you'll find hint like these:
> 
> ```
> [2026-05-28 12:08:12,474] mapproxy.featureinfo - WARNING - Failed to parse content as XML / HTML, returning empty element tree
> [2026-05-28 12:08:12,474] mapproxy.featureinfo - WARNING - Document has no body element, skipping: 
> ```

marcjansen (approved):

ahennr:
Thx for your reviews.

PR #1437, opened 2026-06-08
Fix WMTS demo passing matrix IDs instead of resolutions

ahennr:
WMTS tilegrids use `NamedGridList` consisting of an array of mappings for `matrixId` and `resolution`. Due to typing changes and refactorings, the matrix IDs are passed to the WMTS demo instead of the resolutions, leading to a broken map. This PR fixes this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,7 +16,7 @@ nosetests*.xml
 .pydevproject
 .tox/
 /cache_data/
-
+venv/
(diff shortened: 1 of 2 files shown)


marcjansen (approved):
Thanks, please merge!

hwbllmnn (approved):

PR #1451, opened 2026-06-18
fix(image): don't convert JPEG output back to RGBA mode

### LINKED ISSUE #1446: OSError: cannot write mode RGBA as JPEG when caching transparent WMS source with mixed or jpeg format
## Summary

When a WMS source returns a transparent PNG (`format: image/png`, `transparent: true`) and the cache is configured with `format: mixed` or `format: jpeg`, MapProxy raises `OSError: cannot write mode RGBA as JPEG` during tile creation and storage.

## Environment

- MapProxy 6.1.1 (also reproduced on current `master`)
- Python 3.12
- Pillow ≥ 9.x

## Minimal Reproduction

Any WMS source that returns transparent PNG images, e.g. the public TopPlusOpen or basemap.de WMS:

```yaml
services:
  demo:
  wms:
  wmts:

layers:
  - name: topplusopen
    title: TopPlusOpen
    sources: [topplus_cache]

caches:
  topplus_cache:
    format: mixed
    request_format: image/png
    grids: [webmercator, rlp_grid_25832]
    sources: [wms_topplus_open]
    link_single_color_images: true
    minimize_meta_requests: true
    image:
      mode: RGBA
      colors: 0
      transparent: true
      encoding_options:
        jpeg_quality: 85

sources:
  wms_topplus_open:
    type: wms
    req:
      url: https://sgx.geodatenzentrum.de/wms_topplus_open
      layers: web

grids:
  webmercator:
    base: GLOBAL_WEBMERCATOR
  rlp_grid_25832:
    srs: 'EPSG:25832'
    bbox: [ 250000.0, 5420000.0, 500000.0, 5648000.0 ]
    bbox_srs: 'EPSG:25832'
    origin: 'nw'
    tile_size: [ 512, 512 ]
    res: [560, 280, 140, 70, 28, 14, 7, 2.8, 1.4, 0.7, 0.28, 0.14]

globals:
  cache:
    meta_size: [ 4, 4 ]
    meta_buffer: 80
  image:
    resampling_method: bicubic
    paletted: false
    formats:
      image/jpeg:
        encoding_options:
          jpeg_quality: 80
      image/png:
        format: image/png
        mode: RGBA
```

```sh
mapproxy-util serve-develop my-config.yaml -b 0.0.0.0:8080
curl -s "http://localhost:8080/wmts/topplusopen/rlp_grid_25832/4/7/7.png"
```

## Error

```
File "mapproxy/image/__init__.py", line 426, in img_to_buf
    img.save(buf, format, **defaults)
  File "PIL/JpegImagePlugin.py", line 672, in _save
    raise OSError(msg) from e
OSError: cannot write mode RGBA as JPEG
```

## Root Cause

In `img_to_buf` (`mapproxy/image/__init__.py`), the JPEG branch correctly converts the image to RGB:

```python
if format == 'jpeg':
    img = img.convert('RGB')
```

However, a subsequent block unconditionally re-converts the image based on `image_opts.mode`:

```python
if image_opts.mode is not None and img.mode != image_opts.mode:
    img = img.convert(image_opts.mode)  # reverts RGB → RGBA
```

When `image_opts.mode = 'RGBA'` (propagated from the transparent WMS source), this converts the already-prepared RGB image back to RGBA immediately before `img.save(buf, 'jpeg')`, causing the crash.

## Suggested Fix

Move the JPEG RGB conversion to after the `image_opts.mode` block so it always takes effect last:

```python
if image_opts.mode is not None and img.mode != image_opts.mode:
    img = img.convert(image_opts.mode)

# JPEG does not support alpha — always ensure RGB regardless of mode opts
if format == 'jpeg' and img.mode != 'RGB':
    img = img.convert('RGB')

img.save(buf, format, **defaults)
```

## Workaround

Configure the cache to use `png` format instead of `mixed` or `jpeg`:

```yaml
caches:
  topplusopen_cache:
    format: png
```

gaoflow:
When a WMS source returns RGBA images (e.g., transparent PNG) and the cache uses `format: jpeg` or `format: mixed`, `img_to_buf` converts the image to RGB for JPEG encoding at line 401, but then the generic mode conversion at line 423-424 reverts it back to RGBA because `image_opts.mode` is still `'RGBA'`.

This causes Pillow to raise `OSError: cannot write mode RGBA as JPEG`.

Skip the mode conversion when JPEG format requires RGB.

Fixes #1446

### DIFF
--- a/mapproxy/image/__init__.py
+++ b/mapproxy/image/__init__.py
@@ -421,7 +421,10 @@ def img_to_buf(img: Image.Image, image_opts, georef=None) -> BytesIO:
         del img.info['transparency']
 
     if image_opts.mode is not None and img.mode != image_opts.mode:
-        img = img.convert(image_opts.mode)
+        # Don't convert JPEG output back to RGBA — Pillow cannot write
+        # RGBA as JPEG (fixes #1446).
+        if not (format == 'jpeg' and image_opts.mode == 'RGBA'):
+            img = img.convert(image_opts.mode)
 
     img.save(buf, format, **defaults)
     buf.seek(0)


ahennr:
Thanks for this @gaoflow . It should already be fixed with https://github.com/mapproxy/mapproxy/pull/1447 shouldn't it?

binggao1230:
Yes, #1447 covers the same JPEG/RGBA failure more completely and includes regression tests. Closing this duplicate to keep the queue clean. Thanks for checking.

PR #1453, opened 2026-06-24
installing uswgi through pip eliminates python-linking errors

### LINKED ISSUE #1454: mapproxy 6.1.1: docker-alpine-setup: is it running?
Has anyone tried to get the mapproxy 6.1.1 running with an docker-alpine image? Has it worked?
the versions 6.0.1 & 6.1.0 run like a charm but with 6.1.1 I've always get stuck and get this stacktrace:
```
File "/mapproxy/app.py", line 23, in <module>
from mapproxy.wsgiapp import make_wsgi_app
ModuleNotFoundError: No module named 'mapproxy'
ModuleNotFoundError: No module named 'mapproxy'
unable to load app 0 (mountpoint='') (callable not found or import error)
*** no app loaded. going in full dynamic mode ***
*** uWSGI is running in multiple interpreter mode ***
```
the folder mapproxy exists the file app.py, too.
I've throw ai against it and it says:
uWSGI is linked against Python 3.14, but mapproxy is only installed under Python 3.12 
try to resolve this by installing uwsgi through pip

ffischer1984:
the mapproxy 6.0.1-version works like a charm, I've started to upgrade to mapproxy 6.1.1 and always got this error:
```
File "/mapproxy/app.py", line 23, in <module>
from mapproxy.wsgiapp import make_wsgi_app
ModuleNotFoundError: No module named 'mapproxy'
```
so I think that's a linking problem:
uWSGI is linked against Python 3.14, but mapproxy is only installed under Python 3.12 

try to resolve this by installing uwsgi through pip (to be honest throw ai against the problem and that was there solution)

please handle with care

closes #1454

### DIFF
--- a/Dockerfile-alpine
+++ b/Dockerfile-alpine
@@ -78,11 +78,11 @@ FROM base AS nginx
 # use current version of nginx
 ARG NGINX_VERSION=1.25.3
 
-ENV MAPPROXY_ALPINE=true
+ENV PATH="/mapproxy/.local/bin:${PATH}"
 
 USER root:root
 
-RUN apk --no-cache add ca-certificates uwsgi uwsgi-python3 supervisor pcre-dev && \
+RUN apk --no-cache add ca-certificates supervisor pcre-dev && \
   rm -rf /var/cache/apk/*
 
 RUN apk --no-cache add build-base linux-headers openssl-dev wget zlib-dev && \


ahennr (changes requested):
I can confirm that the current `alpine-nginx` Docker image reproduces the error you described, while `6.1.1-nginx` works correctly, doesn't it?

One thing worth double-checking in your changes: you removed `$MAPPROXY_ALPINE`, which is actually referenced in `docker/run-nginx.sh`. Could you verify that your changes don't inadvertently affect the required uWSGI flags?

Regarding the root cause: 
`apk --no-cache add uwsgi uwsgi-python3` upgrades Python to 3.14, which appears to be a change introduced in the Alpine base image starting with version `3.24` (see also [here](https://pkgs.alpinelinux.org/package/v3.24/main/x86_64/uwsgi-python3)). Alpine `3.23.3` still ships with the expected version, see also https://pkgs.alpinelinux.org/package/v3.23/main/x86_64/uwsgi-python3

As a potential workaround, it might be sufficient (for now) to pin the base-libs image to `python:3.12-alpine3.23` and leave the rest of the `Dockerfile` unchanged:

```Dockerfile
FROM python:3.12-alpine3.23 AS base-libs
...
```

Could you confirm whether this fix resolves the issue on your end as well?

weskamm:
fixed by https://github.com/mapproxy/mapproxy/pull/1304

PR #1455, opened 2026-07-04
Count the partial edge tile in TileGrid.grid_sizes()

gaoflow:
`TileGrid.grid_sizes()` computed the number of tiles per resolution with integer floor division of the pixel span:

```python
x = max(math.ceil(width // res / self.tile_size[0]), 1)
y = max(math.ceil(height // res / self.tile_size[1]), 1)
```

`width // res` floors the pixel count before the division and `ceil`, so when the bbox is not an exact multiple of the tile extent the partial tile covering its edge is dropped and the grid is one tile too small in that dimension. A point inside the bbox but within that edge tile then maps to a tile index that the reported grid size says does not exist.

The fix uses true division and rounds to 6 decimals before `ceil`, so a genuine partial tile is counted while floating-point noise (e.g. from sqrt2 resolutions) does not spuriously add an empty tile:

```python
x = max(math.ceil(round(width / res / self.tile_size[0], 6)), 1)
y = max(math.ceil(round(height / res / self.tile_size[1], 6)), 1)
```

Added tests covering a partial edge tile (projected and geographic), a parametrized check that a point inside the bbox maps to an existing tile, and a sqrt2 grid guard that the count is not inflated.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,3 +1,12 @@
+Unreleased
+~~~~~~~~~~
+
+Fixes:
+
+  - Count the partial edge tile in TileGrid.grid_sizes() so a bbox that is not
+    an exact multiple of the tile extent is fully covered
+
+
 7.0.0 2026-07-17
 ~~~~~~~~~~~~~~~~
 
--- a/mapproxy/grid/tile_grid.py
+++ b/mapproxy/grid/tile_grid.py
@@ -252,8 +252,12 @@ class TileGrid(object):
         height = self.bbox[3] - self.bbox[1]
         grids = []
         for idx, res in self.resolutions.items():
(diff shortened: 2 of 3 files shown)


weskamm:
This seems to be a breaking change? What happens to exisiting caches, will it break them / lead to wrong or missing tiles?

binggao1230:
Not breaking for existing caches: the fix only adds the partial edge tile to the grid extent, so the tile coordinate range grows by at most one column/row at the bbox edge — every existing tile keeps its coordinates and stays cache-valid. What changes is that the edge strip which previously fell outside the tile range now gets a real tile instead of being silently uncovered. Rebased onto master (7.0.0) — CHANGES entry kept as an Unreleased Fixes bullet.

PR #1458, opened 2026-07-16
Upgrade workflows to use python 3.14

weskamm:
This will upgrade the workflows to also use python 3.14.
Shapely and pyproj needed an updated for that.
Docker Images now also use 3.14 (https://github.com/mapproxy/mapproxy/pull/1304)


jansule (approved):
Very nice!

PR #1463, opened 2026-07-17
Sanitize dimension parameters

weskamm:
This fixes a potential path traversal issue in serializing / appending dimension parameters.


jansule (approved):
Very nice! This whole concept still needs an overhaul, though.

weskamm:
Thanks for review, and yes, i do agree!

PR #1470, opened 2026-07-25
fix(s3): normalize S3 tile keys to forward slashes

### LINKED ISSUE #1421: S3Cache generates backslash paths on Windows, breaking both read and write operations on Linux
## Summary

When running MapProxy on Windows (e.g. for seeding via mapproxy-seed), the S3Cache stores and looks up tile objects using backslash-separated S3 keys (e.g. cache_folder\10\4\100.png) instead of the correct forward-slash keys (cache_folder/10/4/100.png).

This breaks all S3 cache operations on Windows, including:

* Writing tiles (seeding)
* Reading tiles (cache lookups during serving, if served from Windows)
* Cache existence checks (is_cached)
* Tile removal

## Root Cause

All tile_location_* functions in mapproxy/cache/path.py build paths using os.path.join():

```python
tile.location = os.path.join(
    cache_dir, dimensions_part(dimensions), level_part(str(z)),
    str(x), str(y) + '.' + file_ext
)
```

On Windows, os.path.join() uses \ as the separator, which is correct for local filesystem paths. However, the tile_key method in mapproxy/cache/s3.py passes this value directly as an S3 object key:

```python
def tile_key(self, tile):
    return self._tile_location(tile, self.base_path, self.file_ext).lstrip('/')
```

S3 object keys have no concept of a filesystem separator — \ is a valid, literal character in an S3 key. So cache_folder\10\4\100.png and cache_folder/10/4/100.png are treated as two completely distinct objects. Tiles seeded on Windows are invisible to any Linux-based MapProxy instance reading from the same bucket, and vice versa.

## Proposed Fix

The fix should be applied in s3.py in the tile_key method, since this is the single point where the filesystem path gets converted to an S3 key. Normalizing the separator here covers all operations (read, write, existence check, delete) without modifying the generic path-building logic in path.py:

```python
def tile_key(self, tile):
    return self._tile_location(tile, self.base_path, self.file_ext).lstrip('/').replace('\\', '/')
```

This ensures that S3 keys always use forward slashes regardless of the operating system MapProxy is running on.

## Steps to Reproduce

1. Install MapProxy on Windows
1. Configure an S3 cache with any directory_layout
1. Run mapproxy-seed — tiles are stored with backslash-separated keys
1. Any read or existence check from a Linux instance (or after the fix) will miss all seeded tiles

## Expected Behavior

S3 object keys must always use / as the path separator, independent of the host OS.

## Environment

MapProxy version: 6.0.1
OS: Windows (any version)
Python: 3.x

Sanjays2402:
Closes #1421

`tile_key()` passed the location built by `os.path.join()` straight through as an S3 object key, so on Windows the key contained backslashes — which S3 treats as literal characters, not separators. Tiles seeded on Windows therefore landed under different keys than the ones any other instance reads.

Backslashes are now normalized to `/` in `tile_key()`, which covers read, write, `is_cached` and removal in one place, leaving the generic path building in `mapproxy/cache/path.py` untouched.

Regression test patches `mapproxy.cache.path.os.path` to `ntpath` to produce a Windows-style location; it fails on master and passes with the fix.

### DIFF
--- a/mapproxy/test/unit/test_cache_s3.py
+++ b/mapproxy/test/unit/test_cache_s3.py
@@ -13,8 +13,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import ntpath
+
 import pytest
 
+import mapproxy.cache.path
 from mapproxy.extent import MapExtent
 from mapproxy.srs import SRS
 
@@ -91,3 +94,12 @@ class TestS3Cache(TileCacheTestBase):
 
         # raises, if key is missing
         boto3.client("s3").head_object(Bucket=self.bucket_name, Key=key)
+
+    def test_tile_key_windows_separator(self, monkeypatch):
+        # on Windows the tile location is built with backslashes, but S3 object
+        # keys must always use forward slashes
+        monkeypatch.setattr(mapproxy.cache.path.os, 'path', ntpath)
+        cache = S3Cache('/mycache/webmercator', 'png', bucket_name=self.bucket_name, directory_layout='tms')
+
+        assert cache.tile_key(self.create_tile((12345, 67890, 12))) == \
+            'mycache/webmercator/12/12345/67890.png'
--- a/mapproxy/cache/s3.py
+++ b/mapproxy/cache/s3.py
@@ -96,7 +96,11 @@ class S3Cache(TileCacheBase):
         return f"https://{self.bucket_name}.s3.{self.region_name}.amazonaws.com/{self.tile_key(tile)}"
 
     def tile_key(self, tile):
-        return self._tile_location(tile, self.base_path, self.file_ext).lstrip('/')
+        # tile locations are built with os.path.join, which uses '\' as separator
+        # on Windows; S3 object keys always use '/' and treat '\' as a literal
+        # character, so keys must be normalized independently of the host OS.
+        location = self._tile_location(tile, self.base_path, self.file_ext)
+        return location.replace('\\', '/').lstrip('/')
 
     def conn(self):
         if boto3 is None:


ahennr (approved):

mapproxy/test/unit/test_cache_s3.py
```diff
@@ -91,3 +94,12 @@ def test_tile_key(self, layout, tile_coord, key):
 
         # raises, if key is missing
         boto3.client("s3").head_object(Bucket=self.bucket_name, Key=key)
+
+    def test_tile_key_windows_separator(self, monkeypatch):
+        # on Windows the tile location is built with backslashes, but S3 object
+        # keys must always use forward slashes
+        monkeypatch.setattr(mapproxy.cache.path.os, 'path', ntpath)
+        cache = S3Cache('/mycache/webmercator', 'png', bucket_name=self.bucket_name, directory_layout='tms')
```
weskamm, line 102: to be honest, i was expecting a test that supplies a windows path and it returns a unix style path, but here i cannot see where you actually test the backslash replacement. Could you please elaborate?

Sanjays2402:
Good question @weskamm — the existing test didn't use a hardcoded Windows path; it simulated Windows by monkeypatching `mapproxy.cache.path.os.path` to `ntpath` so the tile location is built with backslashes, then asserted the resulting key uses forward slashes. That did exercise the replacement, but the input side wasn't visible.

I've made it explicit now: the test asserts the simulated location really contains a backslash (so the normalization path is definitely hit) and that the resulting key contains none, alongside the expected `mycache/webmercator/12/12345/67890.png` value.

PR #1472, opened 2026-08-03
Support for dimensions in mbtiles cache

charlienegri:
a new dedicated class `MBTilesDimensionsCache` is created to support cached mbtiles dimension  
the testing added is only for the cached case , I could use some help in figure out how to build on the existing fixtures for the uncached case if that's desired


weskamm:
Thanks for the contribution. Could you please revert the formatting changes (quotes, linebreaks, ..)?
Most changes of this PR are unrelated and the main changes become hard to find / read.

PR #1480, opened 2026-08-23
docs: document SSL_CERT_FILE/SSL_CERT_DIR support for source CA verification

razbroc:
### What

Documents, in the HTTPS section of `doc/configuration.rst`, that MapProxy honours the standard OpenSSL environment variables `SSL_CERT_FILE` and `SSL_CERT_DIR` for verifying HTTPS sources — and that `REQUESTS_CA_BUNDLE` / `CURL_CA_BUNDLE` do **not** work.

Docs only; no code or behaviour change.

### Why

`build_https_handler()` in `mapproxy/client/http.py` falls back to `ssl.create_default_context()` when `http.ssl_ca_certs` is unset. That call runs `load_default_certs()` -> `set_default_verify_paths()`, so OpenSSL already picks up `SSL_CERT_FILE` and `SSL_CERT_DIR`:

```
$ SSL_CERT_FILE=/etc/ssl/private-ca/root-ca.crt python3 -c \
    "import ssl; print(ssl.get_default_verify_paths().cafile)"
/etc/ssl/private-ca/root-ca.crt
```

This is the natural way to trust a private root CA in a containerized deployment — the CA is mounted into the container and one env var covers every HTTPS source, with no change to `mapproxy.yaml`. But the HTTPS section documented only `ssl_ca_certs`, so it was not discoverable.

The counterpart is the confusing part: `REQUESTS_CA_BUNDLE` is widely assumed to be the generic Python CA knob, but it is only read by the `requests` library. MapProxy makes its source requests through `urllib.request`, so setting it has no effect at all and the connection still fails with a verification error. We hit exactly this in a Kubernetes deployment — `REQUESTS_CA_BUNDLE` was set for the mounted root CA and MapProxy kept rejecting the source until `SSL_CERT_FILE` was set alongside it. The added note should save others that debugging session.

### Verification

`sphinx -b html doc/` builds successfully with no new warnings.


razbroc:
Closing this — opened by mistake, it was meant for a downstream repository. Sorry for the noise.

